PR #54, opened 2024-07-12
Implement Undo

bkrmendy:
## Description

This PR adds undo support to Onlook.

### Architecture
This PR implements the two-stacks approach to undo, where actions are pushed onto an undo stack, which are popped on undo and pushed onto an undo stack. Repeated undos move action from the undo stack to the redo stack. If an action is added to the undo stack, the redo stack is cleared (this is a general UX pattern). 

To deal with inputs where users don't come up with the final value right away, such as color pickers or text inputs, the history system in this PR introduces a concept called transactions. When a transaction is in progress, actions handed to history are not pushed on the undo stack, but are instead stored in the transaction. The action in the transaction is only pushed on the undo stack when a transaction is committed. Undo automatically commits any transactions in progress. 

Each action is described by an object that contains all information to construct an inverse action (see the `Change` type).  `Actions` should be a union type discriminated by `type`, new types should be added as new editor features become available.
 
### Out of scope for this PR
- Redo is not actually implemented in the PR, but its implementation should be analogous to undo.
- Defining more actions
- Cmd/Ctrl-Z should work - since Onlook doesn't support shortcuts yet, this is deferred
 
- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [x] Refactor
- [ ] Other


app/common/history.ts
```diff
@@ -0,0 +1,49 @@
+export interface UpdateStyleAction {
+  type: "update-style";
+  targets: Array<{ webviewId: string; selector: string }>;
+  selector: string;
+  style: string;
+  value: string;
+  originalValue: string;
+}
+
+export type Action = UpdateStyleAction;
```
Kitenite, line 10: Nice I like this, leaves room for growth. We may eventually handle:
1. arbitrary props like src, className, etc. 
2. text 
3. structure (insert element, move element, remove element)

app/src/lib/editor/engine/index.ts
```diff
@@ -3,27 +3,48 @@ import { OverlayManager } from "./overlay";
...
 
     get state() { return this.elementState; }
     get overlay() { return this.overlayManager; }
     get webviews() { return this.webviewManager; }
     get code() { return this.codeManager; }
 
-    updateStyle(style: string, value: string) {
-        this.state.selected.forEach((elementMetadata) => {
+    private updateStyle(targets: UpdateStyleAction['targets'], style: string, value: string) {
+        targets.forEach((elementMetadata) => {
```
Kitenite, line 22: Nice!

app/common/history.ts
```diff
@@ -0,0 +1,49 @@
+export interface UpdateStyleAction {
+  type: "update-style";
+  targets: Array<{ webviewId: string; selector: string }>;
+  selector: string;
+  style: string;
```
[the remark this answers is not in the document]
bkrmendy: I would follow the pattern from this PR https://github.com/onlook-dev/monorepo/blob/main/editor/src/lib/tools/edit/history.ts, making an action for each "kind" of UI action. Obvs the boundaries are fuzzy here, but as a general principle instead of "superactions" that are the general case of many specific actions, I'd go with small and simple actions that are straightforward to understand in isolation, instantiate at a call site and reverse on undo

Kitenite:
>  Cmd/Ctrl-Z should work - this should be implemented on top of a shortcut system that doesn't have to be thrown away later

For reference, here's a [previous implementation](
https://github.com/onlook-dev/monorepo/blob/c4b9120ae212ace084705a634d4eedb6bf58a425/editor/src/lib/tools/hotkeys.ts#L15) of hotkeys here that is planned to port over. Some interesting attributes:
It is fed in the editor engine such that actions are determined by state of the editor engine and swapped based on the currently active tool. 

I was planning on adding a lighter version of this at some point. 

>  "Continuous controls" such as the color picker should only add to the history stack once the user is done with picking a value, but while the color is being tweaked, the changes should still show up interactively

I have 3 proposals for this:

1. We debounce certain continuous inputs such as color such that only 1 action is recorded. 
     a. This should be done in the history tracker, not on the update function in order to keep the UI responsive. 
     b. For example, given a buffer of 1 second, regardless of how many actions are received, we only keep the last action. If an action is received within the buffer, we restart the counter. 
2. We update `updateElementStyle` to take in a `record` field, determining whether the action should go into history. We then handle the debounce at the input level
3. We collapse every event of the same kind to the last event. So regardless of how many event of the same kind goes in consecutively, we only keep 1. 


I dislike option 2 because we will need to handle debouncing events at the input level. I think a smart history system should handle this. 
I dislike option 3 because we want to keep non-continuous events such as alignment even if they're the same time. 

Hence, my recommendation is 1. What are your thoughts @bkrmendy ?

Kitenite:
> "Continuous controls" such as the color picker should only add to the history stack once the user is done with picking a value, but while the color is being tweaked, the changes should still show up interactively

I have 3 proposals for this:

1. We debounce certain continuous inputs such as color such that only 1 action is recorded.
       a. This should be done in the history tracker, not on the update function in order to keep the UI responsive.
       b. For example, given a buffer of 1 second, regardless of how many actions are received, we only keep the last action. If an action is received within the buffer, we restart the counter.
2. We update updateElementStyle to take in a record field, determining whether the action should go into history. We then handle the debounce at the input level
3. We collapse every event of the same kind to the last event. So regardless of how many event of the same kind goes in consecutively, we only keep 1.

I dislike option 2 because we will need to handle debouncing events at the input level. I think a smart history system should handle this.

I dislike option 3 because we want to keep non-continuous events such as alignment even if they're the same time.

Hence, my recommendation is 1. What are your thoughts @bkrmendy ?

Kitenite:
>  All controls should call updateElementStyle with an appropriate Change object - this can be factored out into a helper that constructs the appropriate change object based on the original element style and the value to be applied

I like this. Do you imagine any issues with this for more complex cases such as text changes, inserting element, removing elements, etc.

Kitenite:
> Each action is described by an object that contains all information to construct an inverse action (see the Change type). Actions should be a union type discriminated by type, new types should be added as new editor features become available.

One thing that I really like with this action pattern is that I plan to have agents introduced in the future. You can see an early version of this [here](https://github.com/onlook-dev/monorepo/blob/c4b9120ae212ace084705a634d4eedb6bf58a425/editor/src/lib/components/editor/ChatTab.svelte#L61). 

An agent can have enough context to call an action apply that. I imagine macro actions like rearranging a page or applying a certain color scheme to the page will be feasible in the future.

bkrmendy:
> > All controls should call updateElementStyle with an appropriate Change object - this can be factored out into a helper that constructs the appropriate change object based on the original element style and the value to be applied
> 
> I like this. Do you imagine any issues with this for more complex cases such as text changes, inserting element, removing elements, etc.

The `Change` object is about recording the state that needs to be restored, and it should only depend on the change being made. For example, when updating the text contents of an element, it's impossible to derive the original texts contents from the updated text contents, so we need to record the original as well. In the case of moving an element from one parent to another, the change object would record the original and the updated parents and indices of the moved element.

In the case of adding elements we can skip the Change object altogether, as long as the id/selector of the new element is recorded in the action that describes adding the element (we'll have to figure this out when we get there). On the other hand, the action that removes an element will have to hold on to (some representation of) the original element to be able to restore it. 

All in all I don't see massive complexity ahead, for each action we need to define how to reverse it, but these should be straightforward since in my experience most editing operations already have inverses available independently of the history system.

app/common/history.ts
```diff
@@ -0,0 +1,91 @@
...
+                type: 'update-style',
+                targets: action.targets,
+                style: action.style,
+                change: reverse(action.change),
+            };
+    }
+}
+
+interface Intransaction {
+    type: 'in-transaction';
```
Kitenite, line 32: Nit: I prefer pulling raw strings into constants or enums whenever possible. Mitigates the risk of a typo introducing difficult to trace bugs

app/src/routes/project/EditPanel/ManualTab.tsx
```diff
@@ -85,21 +96,21 @@ const ManualTab = observer(() => {
             return (
                 <NestedInputs
                     elementStyles={elementStyles}
-                    updateElementStyle={updateElementStyle}
+                    updateElementStyle={updateElementStyle(elementStyles[0])} // TODO: this needs a more general approach
```
Kitenite, line 99: Out of scope for this PR but I don't like the updateElementStyle being passed around pattern very much which was done for historical reasons. I will very likely move this into the EditorEngine in a future refactor. 

Same with the way elementStyles are handled and generated. I think it's too complicated. 

Nothing actionable here, just FYI.

app/src/lib/editor/engine/index.ts
```diff
@@ -44,20 +46,49 @@ export class EditorEngine {
...
                 selector: elementMetadata.selector,
-                style,
-                value,
+                style: style,
+                value: value,
             });
         });
     }
 
+    private dispatchAction(action: Action) {
```
Kitenite, line 63: Refactoring notes:
Thoughts on exposing the history object such that the EditorEngine API is cleaner?
`editor.history.runAction vs editor.runAction`

Or alternatively, having an action object that manages actions, transaction state, and history.  `editor.actions.run(action)`. 

Very minor, 2-way door decisions.

app/common/history.ts
bkrmendy: I think Typescript guards against this, but will do

app/src/lib/editor/engine/index.ts
bkrmendy: yeah that sounds good, added it to the todo list

Kitenite:
I gave it a try and it works well! One minor thing is that the editor UI doesn't update when the undo transaction is run. This is because the elementStyles doesn't get refreshed. A simple fix for now is to re-assign the relevant selected element after an action is run. 

Potential block to add this for now:
```
undo() {
        const action = this.history.undo();
        if (action == null) {
            return;
        }

        this.dispatchAction(action);
    }
```
I have a refactoring planned that will fix this more thoroughly

app/src/lib/editor/engine/index.ts
bkrmendy: Pushed it in https://github.com/onlook-dev/studio/pull/54/commits/70f843498e975aa4023737057329a378f6554ba8. I didn't put `runAction` on the new property because it's not history-specific, but I put `*Transaction` and `undo` in there

app/common/history.ts
bkrmendy: implemented in https://github.com/onlook-dev/studio/pull/54/commits/8e0cc0446bbe1b21bf25f24dc73ee793b4000da1

Kitenite (approved):
Awesome work!

PR #69, opened 2024-07-17
Added in first pass at variables

drfarrell:
Consolidated variables from Figma to make naming consistent and start to establish our design system for future features.

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Other


app/tailwind.config.js
```diff
@@ -51,19 +50,196 @@ module.exports = {
...
+                }],
             },
-            borderRadius: {
+
+            /*--- NUMERIC VALUES –--*/
+            borderRadius: { //to be consolidated
                 lg: 'var(--radius)',
                 md: 'calc(var(--radius) - 2px)',
                 sm: 'calc(var(--radius) - 4px)',
+                'x-small': '2px',
```
drfarrell, line 236: Actually I think I'll probably remove these radii – seems a bit over-defined for now.

app/tailwind.config.js
```diff
@@ -51,19 +50,196 @@ module.exports = {
...
                 },
-                surface: {
-                    DEFAULT: '#1a1a1a',
+
+                /*--- TOKENS –--*/
+                bg: {
+                    active: 'var(--bg-tertiary)',
                     brand: {
-                        DEFAULT: 'var(--surface-surface-brand)',
-                        secondary: 'var(--surface-surface-brand-secondary)',
```
drfarrell, line 58: Overdefined the variable for bg – removed all of the surfaces and made them bg (also consistent with figma now)

app/tailwind.config.js
Kitenite: Less is small afterall unless you think it's going to be used a lot
`xs` is the convention for extra small

app/tailwind.config.js
```diff
@@ -51,19 +50,196 @@ module.exports = {
...
+                        DEFAULT: 'var(--color-red)',
```
Kitenite, line 58: We're moving away from red yes?

app/tailwind.config.js
```diff
@@ -51,19 +50,196 @@ module.exports = {
...
+                        secondary: 'var(--color-red-20)',
                     },
+                    default: 'var(--color-gray-100)',
```
Kitenite, line 61: I would capitalize default for convention

Kitenite:
Look ok to me. Be sure to test these out or lmk to test a few before marking this ready. I'll do a final pass once you're ready

drfarrell:
This is to address this issue https://github.com/onlook-dev/studio/issues/68

app/tailwind.config.js
drfarrell: I swapped to teal for now – we'll see how it goes. Would like to find a better brand color but we're rollin' with it

app/tailwind.config.js
drfarrell: I just removed it – Tailwind's border radii are pretty solid

Kitenite:
I updated the variables a little. You can see how they're used from the latest changes @drfarrell 

One nitpicky thing (nit): Tailwind variables generally go from lightest 100 to darkest 1000. What do you think about following this pattern with our design? We're currently doing the opposite. 

I would also encourage running the electron app locally when you add variables to make sure it looks alright.

Screenshot 2024-07-17 at 4 21 48 PM

Kitenite (changes requested):
Added some comments. I think it's ready once the comments are addressed

drfarrell:
> I updated the variables a little. You can see how they're used from the latest changes @drfarrell
> 
> One nitpicky thing (nit): Tailwind variables generally go from lightest 100 to darkest 1000. What do you think about following this pattern with our design? We're currently doing the opposite.
> 
> I would also encourage running the electron app locally when you add variables to make sure it looks alright.
> 
> Screenshot 2024-07-17 at 4 21 48 PM

Sure I'm open to this. The original color flow put 700 as one of the brightest options to match Vercel's approach to high-contrast backgrounds, but I think it'd be too complex to expect people to understand that, so I'll adjust the color spread to be more logical.

Kitenite (approved):
It's working well now. Good to go!

PR #148, opened 2024-08-12
Feature/zoom hotkeys for mouse

bkrmendy:
### Description
This PR adds a number of quality of life features related to zooming and panning as per this issue: https://github.com/onlook-dev/studio/issues/121

Breakdown:
- while the spacebar is pressed, the pan mode is active, releasing the spacebar puts the editor into design mode
- while the middle mouse button is pressed, the pan mode is active (same thing as with the space bar)
- holding the command key on macOS and control on any other platform zooms the canvas on scroll
- cmd/ctrl +/- zooms the canvas in/out. For this feature to work, the main menu bar had to be disabled (since the default shortcuts overrode the custom ones). This commit (https://github.com/onlook-dev/studio/commit/ecc09d93d84eadfd0aae0058d8fa5c691354fb2e) could be split out into it's own PR for completeness' sake

Tech notes:
- re: the zoom shortcut (cmd/ctrl +/-): ideally all shortcuts would be handled in a uniform way, so I'm a bit on the fence about adding it now because it would create some tech debt that'll have to be cleaned up when the said uniform shortcut system is added (even though the linked issue asks for the zoom shortcuts). This PR includes these shortcuts anyway, but they can be easily removed by reverting this commit https://github.com/onlook-dev/studio/commit/883f3ba543c993739bdc6eec481bd55aeecd9386 

- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/routes/project/Canvas/index.tsx
+++ b/app/src/routes/project/Canvas/index.tsx
@@ -67,6 +67,34 @@ const Canvas = ({ children }: { children: ReactNode }) => {
         }
     }, [handleWheel]);
 
+    const handleZoomShortcut = (event: KeyboardEvent) => {
+        if (!isMetaKey(event)) {
+            return;
+        }
+
+        event.preventDefault();
+        event.stopPropagation();
+
+        switch (event.key) {
+            case '0':
+                setScale(1);
+                break;
+            case '=':
+                setScale(scale * 1.2);
+                break;
+            case '-':
+                setScale(scale * 0.8);
+                break;
+        }
+    };
+
+    useEffect(() => {
+        window.addEventListener('keydown', handleZoomShortcut);
+        return () => {
+            window.removeEventListener('keydown', handleZoomShortcut);
+        };
+    }, [handleZoomShortcut]);
+
     useEffect(() => {
         editorEngine.scale = scale;
     }, [position, scale]);
--- a/app/src/routes/project/Canvas/index.tsx
+++ b/app/src/routes/project/Canvas/index.tsx
@@ -1,6 +1,7 @@
 import { ReactNode, useEffect, useRef, useState } from 'react';
 import { useEditorEngine } from '..';
 import PanOverlay from './PanOverlay';
+import { isMetaKey } from '@/lib/utils';
 
 const Canvas = ({ children }: { children: ReactNode }) => {
     const [position, setPosition] = useState({ x: 300, y: 50 });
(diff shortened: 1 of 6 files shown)


Kitenite:
> This PR includes these shortcuts anyway, but they can be easily removed by reverting this commit https://github.com/onlook-dev/studio/commit/883f3ba543c993739bdc6eec481bd55aeecd9386

This is a good note, I will add a note to revert the commit in the hotkeys issue.

app/src/routes/project/Canvas/index.tsx
```diff
@@ -1,18 +1,21 @@
...
+    const [isPanning, setIsPanning] = useState(false);
 
     const editorEngine = useEditorEngine();
     const containerRef = useRef<HTMLDivElement>(null);
     const zoomSensitivity = 0.006;
     const panSensitivity = 0.52;
 
     const handleWheel = (event: WheelEvent) => {
-        if (event.ctrlKey) {
+        if (event.ctrlKey || event.metaKey) {
```
Kitenite, line 18: When pinching on trackpad, ctrlKey is enabled on Mac. I think this is silly as a standard but we need this for zoom on track pad to work.

app/src/routes/project/Canvas/index.tsx
```diff
@@ -62,10 +65,81 @@ const Canvas = ({ children }: { children: ReactNode }) => {
         const div = containerRef.current;
         if (div) {
             div.addEventListener('wheel', handleWheel, { passive: false });
-            return () => div.removeEventListener('wheel', handleWheel);
+            div.addEventListener('mousedown', middleMouseButtonDown);
```
Kitenite, line 68: I move these from the PanOverlay so that this works on the webview as well, not just the area outside the webview

Kitenite:
I'm finding that everything is working without disabling the menu bar. I like having the electron default controls for now for things like CMD+R to reload so I'll add it but will know what the solution is if we run into issues @bkrmendy

app/src/routes/project/TopBar/ModeToggle.tsx
```diff
@@ -1,12 +1,20 @@
...
+import { useEffect, useState } from 'react';
 import { useEditorEngine } from '..';
 
 const ModeToggle = observer(() => {
     const editorEngine = useEditorEngine();
-    const [mode, setMode] = useState<EditorMode>(editorEngine.mode);
+    const [mode, setMode] = useState<EditorMode>(makeDesignMode(editorEngine.mode));
+
+    useEffect(() => {
+        setMode(makeDesignMode(editorEngine.mode));
```
Kitenite, line 12: Update this to handle pan mode changing from interact mode. Now it will update the top bar as well.

PR #161, opened 2024-08-15
Resize the Layer panel 

Sachinsingh2002:
i have made the necessary changes, please review it.

### DIFF
--- a/app/src/routes/project/WebviewArea/ResizeHandles.tsx
+++ b/app/src/routes/project/WebviewArea/ResizeHandles.tsx
@@ -19,6 +19,17 @@ const ResizeHandles = observer(({ webviewSize, setWebviewSize }: ResizeHandlePro
     const editorEngine = useEditorEngine();
     const resizeHandleRef = useRef(null);
 
+    const handleMouseMove = (e: React.MouseEvent) => {
+        const maxWidth = window.innerWidth / 3;
+        let newWidth = e.clientX; 
+
+        if (newWidth > maxWidth) {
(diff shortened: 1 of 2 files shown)


app/src/routes/project/WebviewArea/ResizeHandles.tsx
```diff
@@ -19,6 +19,17 @@ const ResizeHandles = observer(({ webviewSize, setWebviewSize }: ResizeHandlePro
     const editorEngine = useEditorEngine();
     const resizeHandleRef = useRef(null);
 
+    const handleMouseMove = (e: React.MouseEvent) => {
```
Kitenite, line 22: @Sachinsingh2002 this is close but I don't think it's right. The change on 
[app/src/routes/project/LayersPanel/index.tsx](https://github.com/onlook-dev/studio/pull/161/files#diff-f0245d5cb38ab89312972b0120a231f7e191c7938935b2c5dd18f4c3cddbfe08) is correct. 

It would be better to add a resize handle (just a div with some event handlers) on the layers panel that captures mouse down event and does the resizing like you did here. 

This is what this component is referring to (the white handle).  
Screenshot 2024-08-15 at 1 25 15 PM

Kitenite (changes requested):
This is almost there but the change in ResizeHandles should be in LayersPanel in my opinion. Currently, if you resize the webview it will bug out. 

@Sachinsingh2002 let me know if this helps :) Thank you for the PR

PR #181, opened 2024-08-19
Feat: Make overlay purple when element is a component

SkyC0der:
### Description

Made the component outline purple when selected on a canvas using hex code #A020F0

- [x ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (changes requested):
This is in the right direction. However, instead of setting the rect color to purple all the time, we want to make it purple when it is a component, otherwise it should still be red.

You can see similar logic in the layers here where it is purple if it is a component. If there is an instance of the element, that means it is a component.

Screenshot 2024-08-19 at 11 24 51 PM

Color logic, notice purple when instance is not null: https://github.com/onlook-dev/onlook/blob/main/app/src/routes/project/LayersPanel/Tree/TreeNode.tsx#L107

How instance is set: https://github.com/onlook-dev/onlook/blob/main/app/src/routes/project/LayersPanel/Tree/TreeNode.tsx#L36

My recommendation is to determine whether the selected element is a component, then pass that flag into the OverlayManager to change the rects' color.

SkyC0der:
Unable to figure out the logic here :(

SkyC0der:
@Kitenite I'm not sure if I implemented the logic as you requested. Been stuck on the issue. Could you review?

Kitenite:
@SkyC0der hey, it's no problem! I added the infrastructure logic for handling this. 

I realized in the process that:
1. I designed and explained this issue poorly
2. This was much more complex than I had anticipated

Thanks for giving it a try!

Kitenite (approved):

SkyC0der:
Thanks @Kitenite . Going through your implementation and apparently there's a lot for me to learn

Kitenite:
No problem @SkyC0der like I mentioned, this should not have been marked as a "good first issue" on our part 😅. I don't think I would've been able to do it if I didn't know the codebase well already. 

I really appreciate the quickness and the willingness to ask for help!

PR #205, opened 2024-08-23
Fix offset issue and updated electron

Kitenite:

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite:
Seems like in production, the zoom level of the page is the same as Chrome. Some pages with different zoom level can mess with the scaling. What a weird bug. Setting the zoom level of the page to 0 works.

PR #208, opened 2024-08-23
Feat: Close Side Panel 

kartik-raj7:

Added the functionality to close and open Layer and Edit panels

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (changes requested):
I think this is in the right direction. The animation is also very nice. I pointed out a few regressions that this introduces which would need to be fixed.

My recommendation is to keep the existing tabs as unchanged as possible and introduce this hiding functionality 1 level higher. 

Optional recommendation:
I see that the 2 panels share a lot of functionalities. A nice quality of life improvement is to refactor these new functionalities into its own component which then wraps around the panel (i.e. A SidePanel class that takes in a parameter left or right and children). 

You can see something similar in this [right click menu PR](https://github.com/onlook-dev/onlook/commit/95a8b06919eba18df00baf658fac13b849b61ed5#diff-6cbab27c6d17628c54d8a55b86ba6da8515f841772afdf23588b346a14d132f9R224). 

Totally optional :) Nice work so far! It's very close

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -74,17 +78,26 @@ const LayersPanel = observer(() => {
     return (
         <div
             className={clsx(
-                'border min-w-60 max-w-96 bg-black/80 backdrop-blur rounded-tr-xl shadow',
```
Kitenite, line 77: This used to have `min-w-60 max-w-96` which would limit the amount you can resize the panel. This is so that when resized, the panel doesn't become too small or too large. This is a regression.

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -74,17 +78,26 @@ const LayersPanel = observer(() => {
...
+                'fixed left-0 z-50 top-20 transition-width duration-300 opacity-100 bg-black/80',
```
Kitenite, line 81: One last thing and this would be good. The transition duration is messing with the ability to drag the panel size.

Kitenite (changes requested):
This is very close to done! 

One last regression is the ability to resize the panel that I mentioned. There's a delay introduced with the dragging. 

This is a good reason to move the resize logic 1 level higher. There's some transition and mix/min width that is interfering with the resizing. 

Very close thanks for being quick and accommodating :)

Kitenite (approved):
Lovely!

PR #260, opened 2024-08-31
fix: for non-components show text "View element code" on right click

### LINKED ISSUE #251: [Bug] Copy for Right-Click Menu for non-Components
If a div is newly inserted or not an explicit component, the right-click menu should say "View element code" not "View component code" 

image

sidmohanty11:
### Description
When we select a component we have the `instance` information but when we select any non-components it's `undefined`, used that logic to differentiate between elements and components.

Screenshot 2024-08-31 at 6 06 37 PM

Screenshot 2024-08-31 at 6 07 17 PM

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

Fixes #251


Kitenite (approved):
Awesome work!

PR #285, opened 2024-09-02
Component discovery

bkrmendy:
### Description

This PR adds the capability to discover components from a project directory and store them with the path of the file they're in. This will make it possible down the line to insert those components into the code with their respective import statements.

Inserting and removing components is out of scope for this PR, to be added in a follow-up PR.

- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -4,10 +4,62 @@ import { EditorMode } from '@/lib/models';
...
+
+    return (
+        <Button variant="outline" size="sm" className="" onClick={onClick}>
+            Connect to Project
+        </Button>
+    );
+}
+
+// TODO: make this update when `components` changes
+const ComponentsList = ({ components }: { components: ReactComponentDescriptor[] }) => {
```
Kitenite, line 42: For components to subscribe to the useEditorEngine props, they need to be wrapped in observer. This also applies to children components. Each layer should subscribe as needed. 
https://github.com/onlook-dev/onlook/blob/241c7873b2e253b7d6c93bde5b03ccf696eae09b/app/src/routes/project/TopBar/index.tsx#L12

The corresponding subclass in EditorEngine also needs to instrument such that it emits changes: 
https://github.com/onlook-dev/onlook/blob/241c7873b2e253b7d6c93bde5b03ccf696eae09b/app/src/lib/editor/engine/dom/index.ts#L9

app/src/lib/editor/engine/projectinfo/index.ts
```diff
@@ -0,0 +1,16 @@
+import type { ReactComponentDescriptor } from '/electron/main/code/components';
+
+export class ProjectInfoManager {
+    private projectComponents: ReactComponentDescriptor[];
+    constructor() {
+        this.projectComponents = [];
```
Kitenite, line 6: This should call `makeAutoObservable(this, {}); ` to be subscribable

app/src/routes/project/LayersPanel/index.tsx
bkrmendy: fixed in https://github.com/onlook-dev/onlook/pull/285/commits/fee273ac862375ba4503c3ae5844f5f58e1eb009

app/src/lib/editor/engine/projectinfo/index.ts
bkrmendy: fixed in https://github.com/onlook-dev/onlook/pull/285/commits/fee273ac862375ba4503c3ae5844f5f58e1eb009

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -4,10 +4,61 @@ import { EditorMode } from '@/lib/models';
...
 import LayersTab from './LayersTab';
 import { capitalizeFirstLetter } from '/common/helpers';
+import { ReactComponentDescriptor } from '/electron/main/code/components';
+import { MainChannels } from '/common/constants';
+import { Button } from '@/components/ui/button';
+
+function ScanComponentsButton() {
+    const editorEngine = useEditorEngine();
+
+    const onClick = useCallback(async () => {
```
Kitenite, line 18: Note: We're planning to save the project's folder at setup. @drfarrell already has some design drafted up so we won't have to search for the folder at this point. This works in the meantime

app/electron/main/code/components.ts
```diff
@@ -0,0 +1,139 @@
...
+                ...descriptor,
+                sourceFilePath: sourceFile.getFilePath(),
+            });
+        }
+    });
+
+    return exportedComponents;
+}
+
+async function scanDirectory(dir: string): Promise<string[]> {
```
Kitenite, line 109: Question: Should we be ignoring node_modules? Seemingly no because libs are all there. 

But maybe we should be explicitly choosing certain libs because this seems to pick up every single React class in node_modules. Maybe it's just organizing them.

app/electron/main/code/components.ts
```diff
@@ -0,0 +1,139 @@
...
+    const sourceFile = project.addSourceFileAtPath(filePath);
+
+    const exportedComponents: ReactComponentDescriptor[] = [];
+
+    sourceFile.forEachChild((node) => {
+        const descriptor = getReactComponentDescriptor(node);
+        if (descriptor != null) {
+            exportedComponents.push({
+                ...descriptor,
+                sourceFilePath: sourceFile.getFilePath(),
```
Kitenite, line 101: I'm seeing a lot of paths pointing to index.d.ts. I would think that that's not the correct path for importing purposes.

Kitenite (approved):
Overall, this is sick. This seems to be the 1 part 3 from your plan below:

> re: components, I'm planning to do a PoC PR that will have
- component discovery (exported functions where the return value contains a props key, with the help of the Typescript compiler)
- component prop type inference (with the help of the Typescript compiler)
- component insertion. We talked about this with @Kiet at length yesterday, I know layers and write to code are concerns, but I want to create a jumping-off point we can play with


Question: This deviates from the earlier discussion about setting a JSON or other file, is that still planned. If not, how do you see this new approach going UI/UX wise?

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -4,10 +4,61 @@ import { EditorMode } from '@/lib/models';
...
+    }, [editorEngine]);
+
+    return (
+        <Button variant="outline" size="sm" className="" onClick={onClick}>
+            Connect to Project
+        </Button>
+    );
+}
+
+const ComponentsList = observer(({ components }: { components: ReactComponentDescriptor[] }) => {
```
Kitenite, line 41: Style nite: I'd move this into its own component

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -42,7 +93,7 @@ const LayersPanel = observer(() => {
                         <LayersTab />
                     </TabsContent>
                     <TabsContent value={TabValue.COMPONENTS}>
-                        <div className="w-full pt-96 text-center opacity-70">Coming soon</div>{' '}
```
Kitenite, line 45: Maybe leave this as a comment until the feature is ready. Usually would remove but will be toggling this for publishing.

app/electron/main/code/components.ts
bkrmendy: great idea, implemented in https://github.com/onlook-dev/onlook/pull/285/commits/1b73457761da4d3270da7c11235cf5d4df0adcbb

app/electron/main/code/components.ts
bkrmendy: fixed by https://github.com/onlook-dev/onlook/pull/285/commits/1b73457761da4d3270da7c11235cf5d4df0adcbb

app/src/routes/project/LayersPanel/index.tsx
bkrmendy: https://github.com/onlook-dev/onlook/pull/285/commits/9aad568f95b1e612fe70b707487567d7d6d4f13d

app/src/routes/project/LayersPanel/index.tsx
bkrmendy: https://github.com/onlook-dev/onlook/pull/285/commits/37ac194f14a789cf1e6f051897e0ee4723a75d8f

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -4,10 +4,42 @@ import { EditorMode } from '@/lib/models';
...
 import LayersTab from './LayersTab';
 import { capitalizeFirstLetter } from '/common/helpers';
+import { ReactComponentDescriptor } from '/electron/main/code/components';
+import { MainChannels } from '/common/constants';
+import { Button } from '@/components/ui/button';
+import { ComponentsList } from './ComponentsList';
+
+const COMPONENT_DISCOVERY_ENABLED = false;
+
+export function ScanComponentsButton() {
```
Kitenite, line 18: VSCode artifact? This should be in the Components List file or its own file

Kitenite (approved):

PR #297, opened 2024-09-04
Tooltip in the Sidepanel for cut-off layer name Text

kartik-raj7:
### Description

Tooltip for concatenated layer name text

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell:
Wooo excited about this one!

app/src/routes/project/LayersPanel/index.tsx
```diff
@@ -45,7 +45,7 @@ const LayersPanel = observer(() => {
                         <PinLeftIcon />
                     </button>
                 </TabsList>
-                <Separator className="mt-1" />
+                <Separator className="mt-1" id="LayersPanelDiv" />
```
Kitenite, line 48: I would move the id to the tab container since we can use it to calculate the width offset
https://github.com/onlook-dev/onlook/pull/298/commits/09ce88b6e7cc9ecf0f8c24b1dfaa4ad5190777f2#diff-f0245d5cb38ab89312972b0120a231f7e191c7938935b2c5dd18f4c3cddbfe08R49

app/src/routes/project/LayersPanel/Tree/TreeNode.tsx
```diff
@@ -146,21 +156,36 @@ const TreeNode = observer(
...
+                                    : '',
+                            )}
+                        >
+                            {instance?.component
+                                ? instance.component
+                                : node.data.tagName.toLowerCase()}{' '}
+                            {node.data.textContent}
+                        </span>
+                    </TooltipTrigger>
+                    {isOverflowing && (
```
Kitenite, line 180: I would just show the tooltip if there's text content for simplicity
https://github.com/onlook-dev/onlook/pull/298/commits/09ce88b6e7cc9ecf0f8c24b1dfaa4ad5190777f2#diff-eced7d84f55e60e45f2876a4fc40b812bb12b845d5b5cdb4d641a619346d96d1R181

app/src/routes/project/LayersPanel/Tree/TreeNode.tsx
```diff
@@ -146,21 +156,36 @@ const TreeNode = observer(
...
+                        <TooltipPortal container={document.getElementById('LayersPanelDiv')}>
```
Kitenite, line 181: You can use sideOffset to move the tooltip position based on the width of the container. Since the problem is that the node is overflowing, we can calculate the offset as the containerWidth - nodeWidth

https://github.com/onlook-dev/onlook/pull/298/commits/09ce88b6e7cc9ecf0f8c24b1dfaa4ad5190777f2#diff-eced7d84f55e60e45f2876a4fc40b812bb12b845d5b5cdb4d641a619346d96d1R183

Kitenite (commented):
I also would increase the length of the text in textContent so that the tooltip has more information like so

https://github.com/onlook-dev/onlook/pull/298/commits/09ce88b6e7cc9ecf0f8c24b1dfaa4ad5190777f2#diff-66c17179f449d824754ee054672c58937e6bd0d9416f38e76a9c8929b88fc03cL232

Kitenite:
Overall, nice work! You had most of it going, just needed to update the tooltip position.

drfarrell:
My recommendation is to make it a max width of 200px, and restrict it to 4 lines of text before going ...

For example:

**So it should look a little something
like this if you were to see the full
length of text that would occupy a
tooltip of the size you would exp...**

app/src/routes/project/LayersPanel/Tree/TreeNode.tsx
drfarrell: I think we'll need it for every type of layer

image

kartik-raj7:
> My recommendation is to make it a max width of 200px, and restrict it to 4 lines of text before going ...
> 
> For example:
> 
> **So it should look a little something like this if you were to see the full length of text that would occupy a tooltip of the size you would exp...**

image

Does this look correct

Kitenite:
Seems like the bottom padding is a little weird?

Kitenite (approved):
Awesome work!

PR #367, opened 2024-09-16
Add login page

Kitenite:

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


app/src/App.tsx
```diff
@@ -6,15 +6,19 @@ import { ThemeProvider } from './components/theme-provider';
...
 export const useAuthManager = () => useContext(AuthContext);
 
 function App() {
+    const routeManager = useRouteManager();
+
     return (
         <ThemeProvider defaultTheme="dark" storageKey="vite-ui-theme">
             <TooltipProvider>
-                <AppBar />
+                {routeManager.route !== Route.LOGIN && <AppBar />}
```
Kitenite, line 21: No App Bar means no drag... Could make it black for the occasion?

app/src/assets/Dunes_Login.png
Kitenite, line 1: I would reuse the other asset or replace it. Also needs to be compressed. Should be 200kb instead of 2 MB

app/src/routes/login/index.tsx
```diff
@@ -1,5 +1,101 @@
...
+import { GitHubLogoIcon } from '@radix-ui/react-icons';
+import { useState, useEffect } from 'react';
+
 const Login = () => {
-    return <>Login</>;
+    const [lastLoginMethod, setLastLoginMethod] = useState<string | null>(null);
+
+    useEffect(() => {
+        // Retrieve the last login method from localStorage
+        const storedMethod = localStorage.getItem('lastLoginMethod');
```
Kitenite, line 14: Nice stuff but this is not what we do. I'll update it before merge.

Kitenite (commented):
Nice! Please don't merge.

app/src/App.tsx
drfarrell: No drag? For the rest of the app or what do you mean? I don't quite understand.

Kitenite: Try dragging the window without the AppBar. You won't be able to drag it if its disabled. So instead, we can make it transparent or black when route is Login. You don't have to do this, I can update it once we agree on an implementation.

Kitenite: Actually, I updated with what I mean

PR #392, opened 2024-09-22
Maintain the size of the top information bar on the window at various zooms

kartik-raj7:
### Description

Maintain the size of the top information bar on the window at various zooms

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [X] Other


app/src/routes/editor/WebviewArea/BrowserControl.tsx
```diff
@@ -32,6 +34,14 @@ function BrowserControls({
...
     onlookEnabled,
 }: BrowserControlsProps) {
+    const editorEngine = useEditorEngine();
+    const [scale, setScale] = useState(1);
+
+    useEffect(() => {
+        setScale(editorEngine.canvas.scale);
+        console.log('print', editorEngine.canvas.scale);
+        console.log(scale);
+    }, [editorEngine.canvas]);
```
Kitenite, line 44: `editorEngine.canvas.scale` should be the value you're listening to.

app/src/routes/editor/WebviewArea/BrowserControl.tsx
```diff
@@ -166,4 +178,4 @@ function BrowserControls({
     );
 }
 
-export default BrowserControls;
+export default observer(BrowserControls);
```
Kitenite, line 181: Not sure if this works the same way. Generally wrapped around the definition of the function like so 

```
const BrowserControls = observer(({
    webviewRef,
    webviewSrc,
    setWebviewSrc,
    selected,
    hovered,
    setHovered,
    onlookEnabled,
}: BrowserControlsProps) => {
```

Kitenite: Example usage: https://github.com/kartik-raj7/onlook/blob/3767ab8ce4234a70d20f7ab620b1abb5c19586ef/app/src/components/AppBar/index.tsx#L9

app/src/routes/editor/WebviewArea/BrowserControl.tsx
Kitenite: This will detect when canvas is reassigned which should never happen. It prints if you listen to the scale value here instead

```
    useEffect(() => {
        setScale(editorEngine.canvas.scale);
        console.log('print', editorEngine.canvas.scale);
        console.log(scale);
    }, [editorEngine.canvas.scale]);
```

Kitenite (changes requested):

Kitenite (changes requested):
This is definitely the correct direction. 2 things I'm noticing:
1. The correct mouse position is detected for the webview. This is good.
2. The overlay needs to be correctly scaled and positioned given the new location. [This is how the rectangles](https://github.com/onlook-dev/onlook/blob/ef1c253093e45437ca046b72f045d6579675b28d/app/src/lib/editor/engine/overlay/index.ts#L73) are currently scaled. Before, the overlay was positioned right over each webview, now it is ever the entire canvas. The rectangles will need to be scaled accordingly.
3. The browser controls also needs to be correctly scaled and positioned. We will have to maintain a browser control for each frame and position them over the correct frames.

Kitenite:
Closing this since we'll like have to revisit in a fresh PR

PR #430, opened 2024-10-01
[SUP] accounts (init)

hitaspdotnet:
### Description

Account's Tables
User profiles and Organizations

We have 2 types of accounts: users and organizations

- account_registry is used as a repository for all accounts (users & orgs) for better relations and unique names
- users is auth user's profiles to provide more accessibility to user entity. also, we'll use this as personal account for personal projects
- organizations is organized workspace contains members/teams/projects. teams and projects are next part after accounts UI is finished.

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [X] Other


backend/supabase/seed.sql
```diff
@@ -0,0 +1,36 @@
+INSERT INTO auth.users(instance_id, id, aud, "role", email, encrypted_password, email_confirmed_at, raw_app_meta_data, raw_user_meta_data, is_sso_user)
+  VALUES ('00000000-0000-0000-0000-000000000000', gen_random_uuid(), 'authenticated', 'authenticated', 'hitasp@outlook.com', app.generate_token(), now(), '{"provider": "email", "providers": ["email"], "account_name": "hitasp"}', '{"display_name": "Hitasp", "bio": "A Test Account"}', FALSE);
```
Kitenite, line 2: fyi some test data are in here. This is not a concern rn since this won't be use in production but your info is here.  Should probably be removed before merging in.

backend/supabase/seed.sql
```diff
@@ -0,0 +1,36 @@
...
+('demo', FALSE),
+('service', FALSE),
+('info', FALSE),
+('webmaster', FALSE),
+('security', FALSE),
+('installer', FALSE);
+
+BEGIN;
+SELECT
+  app.simulate_login('hitasp@outlook.com');
```
Kitenite, line 33: Same here

backend/supabase/config.toml
```diff
@@ -1,202 +1,35 @@
...
-# smtp_port = 54325
-# pop3_port = 54326
-
-[storage]
-enabled = true
-# The maximum file size allowed (e.g. "5MB", "500KB").
-file_size_limit = "50MiB"
 
-[storage.image_transformation]
+[auth.external.github]
```
Kitenite, line 21: Nicely done, great inference on this without knowing our actual setup

backend/supabase/migrations/20240414161927_users.sql
```diff
@@ -0,0 +1,276 @@
...
+        END
+    WHERE
+        u.id = update_user.user_id;
+    RETURN get_user_by_id(update_user.user_id);
+END;
+$$;
+
+GRANT EXECUTE ON FUNCTION public.update_user(uuid, text, text, jsonb, boolean) TO authenticated, service_role;
+
+CREATE OR REPLACE FUNCTION public.search_users(account_name citext DEFAULT NULL)
```
Kitenite, line 260: This allows for searching all users. And is public. This allows a full scan of all user. My understanding is that this will be restricted based on RLS @hitaspdotnet ?

backend/supabase/migrations/20240414161927_users.sql
```diff
@@ -0,0 +1,276 @@
...
+FROM
+    users u
+WHERE
+    u.id = auth.uid();
+END;
+$$;
+
+GRANT EXECUTE ON FUNCTION public.get_me() TO authenticated;
+
+CREATE OR REPLACE FUNCTION public.update_user(user_id uuid, display_name text DEFAULT NULL, bio text DEFAULT NULL, public_metadata jsonb DEFAULT NULL, replace_metadata boolean DEFAULT FALSE)
```
Kitenite, line 231: We'd probably want to restrict this access to only the authenticated user. Would this be automatically restricted by row level security?

Kitenite (commented):
Great work on this. Some questions but otherwise looking good!

backend/supabase/seed.sql
hitaspdotnet: yes, it's for testing triggers. I forgot to delete.

backend/supabase/migrations/20240414161927_users.sql
hitaspdotnet: based on `users_select_policy` all authenticated users can select each other. just select. same as when you search a user in github to send invitation. 
I'll change return type to returns only account_name and ID or a view when views was ready.

backend/supabase/migrations/20240414161927_users.sql
hitaspdotnet: yes,
https://github.com/onlook-dev/onlook/pull/430/files#diff-a94914aeab25d03ab188b8d6032dce5c62168026eae485c707805d1d22143ea0R140-R142

hitaspdotnet: all functions have security of type `DEFINER` and all table RLS are enabled per row and column for insert/update/delete. however, these are needing more security check of course.

PR #461, opened 2024-10-04
[SUP] database queries

hitaspdotnet:
### Description

wrappers for supabase queries and RPC functions

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [X] Other


Kitenite (approved):
Looks good, thanks for the refactor. 
Common needed an alias in vite config so data might need the same. It seem to build just fine so I wouldn't worry about it. Just FYI if something goes wrong on the client. 

https://github.com/hitaspdotnet/onlook/blob/a86b75e8b8803379b3839ba3c28d8307982c5a46/app/vite.config.ts#L21

Issue #307, opened 2024-09-05
[bug] The main window starts off-center on startup


drfarrell:
The window ideally starts in the center of the screen when the app loads, not like the screenshot below

<img width="1510" alt="image" src="https://github.com/user-attachments/assets/2723ee2a-4e95-4684-9a8d-c2ce724da9a0">

ahp64:
Can I take this issue?

Kitenite:
@ahp64 go for it!

stephenm7777:
I can take this if you're cool with it

Kitenite:
For sure @stephenm7777 go for it. It might not be very straightforward but happy to talk through a solution

Kitenite:
@stephenm7777 , how's it going? Do you need any support on this?

ahp64:
Hi, I've been busy the past few weeks but if the issue is still unresolved, I'm more than willing to take it back.

Kitenite:
Back at it again, go for it @ahp64 !

Kitenite:
Merged @ahp64! Nice work. Shouted you out in the Discord as well :D

PR #462, opened 2024-10-05
webview issue

ahp64:
Related to issue #307 
Previously, the react "window" inside the app, done with a <webview>, would be off center when onlook was quit with it off center and then reopened. This was because position was only being reset once the app was fully quit and the cache cleared. Now, the react "window" is reset each time onlook is reopened. I did not address the fact that scaling persists through closing and opening onlook because this was not what the issue was about, but changing that would require minimal edits right next to where my edits were made.

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (approved):
Nice, very elegant solution! 
I will move the logic into editor engine in a separate PR. This works great!

Issue #278, opened 2024-09-02
[Bug] Holding shift occasionally outlines elements in the UI 


drfarrell:
Holding down shift and clicking on elements in the UI occasionally adds a white border around those elements. Holding shift and clicking should not outline any element in the Onlook UI.

<img width="343" alt="image" src="https://github.com/user-attachments/assets/2b6b9692-d8f7-4a62-b3e6-7b6663db3b77">

Moshyfawn:
This is still the default accessibility behaviour of the browser _(Chromium specific)_, which might be better disabled by default and only enabled on specific interactive elements where it makes sense, as Onlook is a desktop application where focus is handled differently to web platforms.

drfarrell:
Ah thanks for the explanation! I didn't know what it was.

abhiroopc84:
Hey @drfarrell, I'd like to work on this, could you assign it to me?

Kitenite:
All yours @abhiroopc84!

drfarrell:
Seems like it works! Thanks @abhiroopc84

PR #463, opened 2024-10-05
Fix: Prevent white outline when holding shift and clicking on UI elements

abhiroopc84:
### Description

This PR addresses issue #278, where holding down the Shift key and clicking on elements in the UI occasionally adds a white border around those elements. This behavior is unintended and should not occur in the Onlook UI.

The issue occurs because of the `:focus-visible` pseudo-class which applies while an element matches the `:focus` pseudo-class and the UA determines via heuristics that the focus should be made evident on the element. Many browsers show a "focus ring" by default in this case, which is removed in this PR.

Although this fixes the issue of unintended white outline on shift click, it also removes the white outline on tab click which makes it harder to navigate the UI using the keyboard.

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/index.css
+++ b/app/src/index.css
@@ -89,3 +89,8 @@ body,
     width: 100%;
     overflow: auto;
 }
+
+*:focus-visible {
+  outline: none !important;
+  box-shadow: unset !important;
+}


drfarrell (commented):
Seems like it got the job done 👍

PR #542, opened 2024-10-15
Make settings button more subtle

Kitenite:

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell:
The context here is that the exit button needs to balance the title text on the left side but using a cross icon may imply "delete" or "clear settings". 

The button should therefore include a label but the current design is a unique instance of the button component that isn't used anywhere else in the app. 

The goal would be to use a standard button that is explicitly saying what the action will be while visually balancing the title

Kitenite:
Won't fix

PR #555, opened 2024-10-16
Command to run a new project on Windows 

### LINKED ISSUE #547: [bug] Creating new project on Windows can't cd absolute path
#### Describe the bug
when the project folder is on windows, D: drive, you can't simply cd to the location.. 

So you have to:
`D: && cd D:\Users\...\onlookdemo && npm run dev`

This should be handled here 
https://github.com/onlook-dev/onlook/blob/main/app/src/routes/projects/ProjectsTab/Create/New/Run.tsx#L22

Utzig26:
### Description
Fix #547 
- just add the [`/d` param](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cd#parameters) to cd command in windows.  
  - > `/d` Changes the current drive as well as the current directory for a drive.

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/routes/projects/ProjectsTab/Create/New/Run.tsx
+++ b/app/src/routes/projects/ProjectsTab/Create/New/Run.tsx
@@ -19,7 +19,9 @@ export const NewRunProject = ({
 }) => {
     const [isRunning, setIsRunning] = useState<boolean>(false);
     const [hasCopied, setHasCopied] = useState<boolean>(false);
-    const codeContent = `cd ${projectData.folderPath} && npm run dev`;
+    
+    const platformCommand = process.platform === 'win32' ? 'cd /d' : 'cd';
+    const codeContent = `${platformCommand} ${projectData.folderPath} && npm run dev`;
 
     function copyToClipboard(text: string) {
         navigator.clipboard.writeText(text);


app/src/routes/projects/ProjectsTab/Create/New/Run.tsx
```diff
@@ -19,7 +19,9 @@ export const NewRunProject = ({
 }) => {
     const [isRunning, setIsRunning] = useState<boolean>(false);
     const [hasCopied, setHasCopied] = useState<boolean>(false);
-    const codeContent = `cd ${projectData.folderPath} && npm run dev`;
+    
+    const platformCommand = process.platform === 'win32' ? 'cd /d' : 'cd';
```
Kitenite, line 23: What happens when the user uses a different drive than D? Should we detect the drive and split it into a separate cd call like you have here?

Utzig26: The `/d` parameter, switches to any disk specified in the `projectData.folderPath` see [here](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cd#parameters)

For example
```shell
C:\users\any> cd \d F:\any\dir\you\want
F:\any\dir\you\want> cd \d F:\any\dir\you\want\same\disk
F:\any\dir\you\want\same\disk>
```

Kitenite: Gotcha, love it! I'll test on my Windows machine and merge. Nice work!

Kitenite: It works great!

PR #556, opened 2024-10-16
fixes #554: fix delete card in light mode

### LINKED ISSUE #554: [bug] Wrong background color in delete item pop-up.
Color styling in lightmode needs to be fixed. You can refer to the project setup cards for the right tokens needed for pop-ups to be compliant with dark / light mode. bg-background-onlook is likely what needs to be used here.

image

aryanbroy:
### Description

Fixes the background color of delete card in light mode

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/components/ui/alert-dialog.tsx
+++ b/app/src/components/ui/alert-dialog.tsx
@@ -34,7 +34,7 @@ const AlertDialogContent = React.forwardRef<
         <AlertDialogPrimitive.Content
             ref={ref}
             className={cn(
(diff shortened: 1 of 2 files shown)


Kitenite (commented):
This seems to work. Can we move this classname to the root AlertDialogContent component instead? This would fix the other instances as well
https://github.com/aryanbroy/onlook/blob/c189054140feef5bd66c302c5a9a726a39681898/app/src/components/ui/alert-dialog.tsx#L34

Issue #533, opened 2024-10-15
[feat] Swipe-animate the project setup flow using Framer Motion


drfarrell:
Using the lessons from https://animations.dev/, create reasonable, smooth animations for the steps that need to happen in the project setup flow.

![Export-1728952422776](https://github.com/user-attachments/assets/bcef62b3-001a-4349-a171-7edf1ed95ea3)

This likely requires a new component type with the animations. The animations should animate the height of the card, the animation from left-to-right, and the animation from right-to-left (animate back a step).

0xsommer:
I am going to work on that @drfarrell

0xsommer:
What do you think @drfarrell?


https://github.com/user-attachments/assets/ff62b502-d64c-4408-9d72-5c250218a6c6

Issue #210, opened 2024-08-23
[FEAT] Adopt Family's style of text transition for button labels


drfarrell:
Inspired by 
https://x.com/wickedmishra/status/1826650539876384914

![Export-1724443551585](https://github.com/user-attachments/assets/272a8b46-396c-41c3-978f-68c6e9511638)

An example is built here: 
https://codesandbox.io/p/sandbox/framer-motion-familys-inline-text-transformation-7qmpjv?file=%2Fsrc%2FApp.tsx

drfarrell:
It is also recommended that for multi-word labels the words that persist between states remain in-tact. 

This tweet has an example: 

https://x.com/mikesommersby/status/1826944977147965775

drfarrell:
A great place for this would be the text of this button, to start 
<img width="266" alt="image" src="https://github.com/user-attachments/assets/b906beb9-ebbf-49a1-84bc-a64e78f56932">

drfarrell:
This would be a great spot to do the button label transition
![Export-1729972755591](https://github.com/user-attachments/assets/c6ad5091-b902-4db2-96d4-62d5c548f852)

PR #559, opened 2024-10-16
#533 [feat] Swipe-animate the project setup flow using Framer Motion

0xsommer:
### Description

Issue: #533 

Adds a directional aware entry and exit animation to the steps of the create project onboarding. The implementation uses Framer Motion and animates the height of the card after measuring its content height in order to achieve a fluid movement.

– A new component called MotionCard was added that can be used from now on to achieve similar card animations in the future.

– The structure of the step components was changed in order to split the MotionCardFooter from the MotionCardContent and make the MotionCardFooter persistent across all the steps.

– LayoutAnimations were added to transition from MotionCardContent state changes, e.g. from the "Choose a folder" button to the representation of the selected folder. 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

https://github.com/user-attachments/assets/51c401fd-454b-472b-8313-c99885894ec8


drfarrell:
Some feedback – 

Seems like the border of the card is slightly shifted over one pixel to the left and one pixel up
image

The changing height animation is so 👌 between card states. 

Totally just a discussion / design / taste point, but what do you think about the text-swiping in? Should we go for something more like #210 where the text elements kind of blend into eachother as they change states so they don't have to travel as far into the screen? Curious your thoughts. It may be a "feel" thing that requires implementation.

drfarrell:
Here's an over-exaggerated version. I think the blend mode is plus-lighter, but I just did a bg-white/30 or something because I couldn't figure out how to layer the blend mode correctly. You'll want to use a bg-foreground or bg-background to comply with lightmode / darkmode, but you can use your own divider (bg-foreground/30, bg-foreground/90, etc.) depending on what you think works best for light and dark mode. You may also need to define different dividers for dark:bg-foreground vs just the regular bg-foreground.

image

I do remember having to do some tricky stuff to get the border on the outside at the right size (0.5px)... I think this is the code on the card.tsx file:

`const Card = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(
    ({ className, ...props }, ref) => (
        <div
            ref={ref}
            className={cn(
                'rounded-xl backdrop-blur-xl bg-black/60 shadow-md text-card-foreground relative',
                'before:absolute before:inset-[-0.5px] before:rounded-[12.5px] before:border-[0.5px] before:border-white/20 before:pointer-events-none',
                className,
            )}
            {...props}
        >
            {props.children}
        </div>
    ),
);
Card.displayName = 'Card';`


As for the animation – thanks for the explanation. I think it looks very cool!

0xsommer:
Okay, understood @drfarrell – thanks for the visual to explain the idea. I changed the box shadow to match the example above and used the variables to adjust according to dark/light settings.

Kitenite:
@0xsommer , it looks good from my testing. I merged from main to fix the conflict. One last item is to run `npm run lint` which will return the below error at some of the new component instance as they are React components that are assigned anonymous functions. 

`Component definition is missing display name  react/display-name`

Otherwise look good to merge! Let me know if you'd rather I add the fix instead :)

0xsommer:
If you can fix it, I'd appreciate it - I'll be away from my computer for a few hours :)

Kitenite (changes requested):
Looks like there's some regression happening for some of the functionalities (namely the  skip ability in import). That would need to be fixed before merging this in.

Kitenite:
Working well, I'll @drfarrell for one last regression test before merging. Nice work here!

Kitenite (approved):
Looking good on my side

Issue #549, opened 2024-10-15
[feat] "Enter to apply" helper text and icon for Tailwind input boxes.


drfarrell:
"Enter to apply" text and icon that shows while the Tailwind box is focused. When the Tailwind input box is not focused, hide the helper text.
<img width="626" alt="image" src="https://github.com/user-attachments/assets/7c15d2d3-f576-4027-abeb-f137fd621ebb">

Ideally it doesn't overlap with the input text, so it may require some margin above it.

Here's the custom return icon .svg
![Return](https://github.com/user-attachments/assets/1c07a8f8-38be-45ff-9cfe-63cffa95aabc)

cduarte3:
Hello @drfarrell , could you please assign me this issue?

Kitenite:
Hey @cduarte3, go for it!

cduarte3:
# Quick update

## Resizing

With the ability to resize that textArea while inputting content, there tends to be some overlap if a user makes the window shorter. This resizing ability can be kept if the "enter to apply" is placed underneath the textArea. 

**Example of placement with keeping ability to resize:**

![image](https://github.com/user-attachments/assets/3bb8cfdb-d53f-482f-b74a-bdd9fdd72d2c)


## My fix so far

So far, I have been able to implement the desired look as requested, just without that user resizing feature. I have made the textArea resize automatically so that there is always margin between the content and the "enter to apply" message.

**My changes:**

![image](https://github.com/user-attachments/assets/0c18af57-83f1-49dd-b847-6bf926f2b95e) ![image](https://github.com/user-attachments/assets/37bc303d-b562-4bea-9cce-f94192cdb2c5)


## Overall

Please let me know if my current implementation is a good idea, or if I should employ the strategy mentioned prior to this. I would think the automatic resize is better for the requested change but I am not sure if you would like to keep the user's ability to resize this textArea. Please let me know of the preference.

Kitenite:
@cduarte3  I like the auto-resize. We can probably add a max line if it gets too long. 

I think we'd still like users to be able to resize. We can implement custom drag areas on the edge for that if the default resizer doesn't work. But I think it's worth opening a PR with what you have currently :)

Kitenite:
Merged, nice work @cduarte3 !

drfarrell:
Thanks for your work on this @cduarte3 it looks slick!

PR #572, opened 2024-10-17
Text area focus hint

cduarte3:
# Description

As per issue https://github.com/onlook-dev/onlook/issues/549, the 'textArea' where a user can input tailwind classes for an object needed a hint or message at the bottom saying "enter to apply" when the object is in focus. This is done to let the user know that while in focus, changes may not be saved, and to save changes they must click the enter key.

I was provided with that piece of information, as well as an SVG for an arrow image, and a reference picture: 

image

My current implementation looks as so:

image

I was able to implement this feature, albeit with some minor changes to the textArea.  Currently it is being auto resized with the user's text and some extra space as the sizing option.

image

Was informed by @Kitenite that users should still be able to resize the box and this can be done so eventually with custom drag areas. However, at the moment I was only able to keep the resize with the text outside of the textArea and in this position:

image

Please let me know if you have any issues with this request or any changes are needed.

# What is the purpose of this pull request? 

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [X] Other

### DIFF
--- a/app/src/routes/editor/EditPanel/inputs/single/TailwindInput.tsx
+++ b/app/src/routes/editor/EditPanel/inputs/single/TailwindInput.tsx
@@ -1,8 +1,9 @@
 import { useEditorEngine } from '@/components/Context';
 import { Textarea } from '@/components/ui/textarea';
 import { sendAnalytics } from '@/lib/utils';
+import { ResetIcon } from '@radix-ui/react-icons';
 import { observer } from 'mobx-react-lite';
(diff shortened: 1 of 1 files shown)


Kitenite (approved):
Nice work this was more complex than meets the eye! I just replaced the image with our icon :)

cduarte3:
@Kitenite Thanks for the update

PR #576, opened 2024-10-17
Added Youtube link during Setup

aPassie:
Add in a button that navigates the user to https://youtu.be/RjFBUkVfy1E?t=45&feature=shared in their browser


drfarrell (changes requested):
Great! Just a few small things to make it feel more on-brand.

The classes need to comply with our tailwind classes so it shows up correctly in lightmode. This should be easy, use classes like – bg-background-onlook or bg-foreground for example.
image

The entire button should be clickable, not just the link. It should hover similarly to the "Click to select folder" button. 

The typography should use text-RegularPlus class for the title and text-Regular class for the a text, just so we keep it within our typography system.

The Icons we use are Radix icons. Please swap the external link icon from lucide to radix – you can just add in the icon into the radix import at the top.

Let's make the dimensions of the photo the correct dimensions so it fits nicely without the top and bottom black bars.

We handle opening external links a little differently than what was proposed... try following what we did for the signin page 
Screenshot 2024-10-17 at 5 35 39 PM

Kitenite (approved):
Look great thanks for this!

PR #577, opened 2024-10-17
Windows Acrylic Background

abhiroopc84:
### Description

This PR includes background changes that make the acrylic background possible on Windows. It might work on MacOS as well if there are no unintended side-effects.
Screenshot (476)
Screenshot (477)
 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell (changes requested):
**Feedback on the acrylic:** 
WOAH this looks sick! It works on Mac too. It's very subtle which is so nice. Super cool!

**Feedback on the layers panel** 
- It seems like the layers panel background is too transperant. Maybe things were changed when you added in the acrylic effect, but it's hard to read the text on the sidepanel now. See below.
- It seems like the layer's have an abupt ending depending on the length of the string or the string continues too far. See the screenshot below:
image
- The Icons look really good on light and dark. One thing for specifically selected classes, is that the icons also need to be white / active or whatever the class is along with the text. See if you can add that class in for that icon.

Screenshot 2024-10-17 at 5 46 33 PM

abhiroopc84:
I'll have a look at the layers panel background and see why it's abruptly ending as well. 

As for the icons, I was wondering how I should style the header icons because the H and the layer number are coloured differently so, when the icon is active, do I colour both the H and the number active with the same colour?

drfarrell:
> I'll have a look at the layers panel background and see why it's abruptly ending as well.
> 
> As for the icons, I was wondering how I should style the header icons because the H and the layer number are coloured differently so, when the icon is active, do I colour both the H and the number active with the same colour?

I think for the case when they are selected they can both be the same color (white/active) just for maximum contrast against the color background. When they aren't selected, I like the half-tone coloring, but you tell me if that's too tough to engineer.

abhiroopc84:
@drfarrell I've fixed both, the issue where in the layers panel background was too transparent and also that of layers having an abrupt ending depending on the length of the string.

Kitenite:
Performance didn't seem to degrade on Mac though I can't tell on my version. Otherwise looks good. Deferring to @drfarrell  for the final review.

Kitenite:
@drfarrell reminder to look at this

Kitenite:
Just a regression test. Still working well after merging from main. There's some UI changes that will require @drfarrell to test/address

Screenshot 2024-10-26 at 1 35 52 PM

Kitenite:
I can also see the difference on my Mac now after updating. Liking it so far.

Kitenite:
Yeah it's a matter of styling at this point while preserving the seethrough-ness of the canvas. Instead of bg-non I'm testing with lower opacity of the original colors which is looking nice but I'll leave it to @drfarrell for this.

Kitenite:
Gonna have to close this since I don't think we can support this while keeping existing styling/behavior

Issue #581, opened 2024-10-17
[feat] Make the Zed logo black when the app is in light mode.


drfarrell:
Make the logo black for light mode and light for darkmode. 

This should be similar to #534 

<img width="298" alt="image" src="https://github.com/user-attachments/assets/b0a367e2-5d84-4a1e-8158-ae7c2fa1777f">

Utzig26:
I worked on #534, I could work on this one too

Yash-007:
I can also work on this issue.

Utzig26:
Hey @drfarrell, both the VSCode and Cursor icons use a silver palette that works great in both light and dark modes. By changing the Zed logo to a similar idea, we achieve the following result:
| Dark | Light |
|--------|--------|
| ![image](https://github.com/user-attachments/assets/59866540-eb8b-4cba-a3bf-75a2169334a8) | ![image](https://github.com/user-attachments/assets/cb77d0f0-4910-481d-aa0d-8c7ccfd6780e) | 

I just changed the .svg
| From this | To this |
|--------|--------|
| ![image](https://github.com/user-attachments/assets/640e2306-00ea-40f9-8a28-9266162fa31f) | ![image](https://github.com/user-attachments/assets/3c6a0484-2fb6-467c-b8d8-99b084489b25)| 

It just took me a few minutes before I tried to write some code. If this is a valid solution, I can create a PR with the new SVG.

Kitenite:
@Utzig26 please do! There's an existing PR as well but I like this clean solution :)

drfarrell:
This seems to be solved #601

PR #583, opened 2024-10-17
fix: Make the Zed logo black in light mode (#581)

GooglyBlox:
### Description

This PR resolves the issue of the Zed IDE SVG not changing colors with the chosen theme. I opted to use simple CSS styling to resolve the issue, as the SVGs for Cursor and VSC don't actually have inherent color-changing by theme (they are just colored natively with enough contrast to appear in dark and light modes). If you're looking for a fix with a little bit more complexity, I can re-do how the IDE SVGs are rendered!

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


app/src/index.css
```diff
@@ -123,6 +123,10 @@
 
         --input: 12 6.5% 15.1%; /* shadcn */
         --ring: 24 5.7% 82.9%; /* shadcn */
+
+        img[src*='zed.svg'] {
```
Kitenite, line 127: Is there a way to scope this filter into the SVG so that it's not in our global styles @GooglyBlox ?

GooglyBlox:
My solution fixes this issue, but it might actually make more sense to use @Utzig26's solution, since they made all of the SVGs actually match in color! It'd probably be better than a hacky filter solution.

Kitenite:
Thanks for the good try anyway @GooglyBlox , I was arriving at the same solution :)

Issue #550, opened 2024-10-16
[feat] UI for our AI Chat – no functionality required


drfarrell:
Implement the following design and its components in the "Ai Styles" Placeholder section of the right properties panel.

<img width="671" alt="image" src="https://github.com/user-attachments/assets/1fa2a160-3cb6-4bf3-b617-30231a65ee7c">

Here's a more detailed screenshot:
<img width="796" alt="image" src="https://github.com/user-attachments/assets/2c16c9b6-5a4e-4949-b2b5-6f91a3ca64fc">

This is the design for the "chat history" button. The "chat history" button should have a hover state label that says "Chat History" and be in an "active" state when the side panel is open. It would be ideal if there was a search bar in the top panel, but that can be added in a later PR. It is OK if the width of the chat history is a bit larger that the properties panel. The "Plus" button is a "New Chat" button and should have a hover state label on it too.

<img width="987" alt="image" src="https://github.com/user-attachments/assets/8d972016-3cdc-4ba5-8cb6-37efa92cb72c">

Please use the proper tokens for foreground / background, as our app is in light / dark mode now.

Also use proper text-sizing tokens from our app tokens indicated by this chart below. Use your best judgement – we can clean up the styles easily if there are issues.
<img width="730" alt="image" src="https://github.com/user-attachments/assets/d65a1301-bc77-4063-82d6-42ba26e3d566">

tanmaytare:
i would like to work on it can you assign it to me

abhiroopc84:
@Kitenite @drfarrell could you assign this to me?

Kitenite:
👍  @abhiroopc84

PR #587, opened 2024-10-18
Feature - UI for AI Chat

abhiroopc84:
### Description

Implemented the AI Chat UI as mentioned in #550. I've increased the width of the EditPanel to accommodate the UI elements better, we can decide on a default width once the resizable EditPanel implementation is done.

image
image

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell:
I did an initial pass through just tidying some stuff up, but I think we'll have a better idea of what the next phase of this will be when we have some of the first AI features together. @Kitenite what do you think?

abhiroopc84:
Which model are you planning to use for the AI features @drfarrell ?

Kitenite:
@abhiroopc84 , I've used openai in the past and their API is always improving so likely that. I personally use claude so will give the docs a look before I lock down to one.

abhiroopc84:
@Kitenite are you planning on fine tuning the model to better perform for our use case? Last I checked, I think Claude does support fine tuning their model, but OpenAI didn't yet support it.

Kitenite:
@abhiroopc84 No plans as of yet. We'll start with some baseline performance, gather some feedback and decide how we want to proceed. There are a lot of different use cases we want to try before optimizing any particular one.

Kitenite (approved):
Oh man I'm excited for this! Nice work

Issue #276, opened 2024-09-02
[FEAT] Style layers and their parent group


drfarrell:
As an enhancement to the UI, it would be great if we could better designate a selected parent group. Ideally, the layers feel part of a rounded-corner group with the selected layer as the brightest. 

The specific tailwind classes are listed in the annotated mockups below.

![image](https://github.com/user-attachments/assets/b8be8adf-23f5-4437-bb36-a975ed650b13)

![image](https://github.com/user-attachments/assets/0be2770e-c25b-4055-90b8-603d07138845)

![image](https://github.com/user-attachments/assets/7a612984-dc2b-43f4-90d0-0033196e2e09)

drfarrell:
<img width="319" alt="image" src="https://github.com/user-attachments/assets/d64f2ba9-6b6b-4e33-bf4c-68fc0d9243d7">

An example of how this would look for layer text that is close to the right side of the panel.

abhiroopc84:
@drfarrell can I work on this?

Kitenite:
@abhiroopc84 have at it! 
Here's where the layers tree lives
https://github.com/onlook-dev/onlook/blob/983c7907fe42f626ea58148fb1c7228e7f4e95d5/app/src/routes/editor/LayersPanel/LayersTab.tsx#L89

This is the layers library we're using https://github.com/brimdata/react-arborist

abhiroopc84:
@Kitenite anything left to be worked on in this issue?

Kitenite:
Looks good I'll close it. @drfarrell fyi in case it should be reopened.

PR #598, opened 2024-10-18
Style parent group

abhiroopc84:
### Description

Added colors to better designate a selected parent group and make the layers feel part of a rounded-corner group with the selected layer as the brightest. #276 
image
image
image
image

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (approved):
Looks like a choice was made to disable horizontal scrolling here but I do prefer it this way. We might try to handle super deep layer tree later.

Nice work!

PR #607, opened 2024-10-19
feat: Animate forward button into view if going forward is possible

### LINKED ISSUE #580: [feat] hide the upper nav-bar "forward" button if there's nothing to go forward to
Ideally this button would animate-into view if there's an option to go forward.
image
Screenshot 2024-10-17 at 5 10 31 PM

j-mahapatra:
### Description

fixes #580

**Changes**: 
- Adds check for forward button to show only if going forward is possible.
- Adds animation when forward button comes into view.

- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/routes/editor/WebviewArea/BrowserControl.tsx
+++ b/app/src/routes/editor/WebviewArea/BrowserControl.tsx
@@ -16,7 +16,6 @@ import {
 } from '@radix-ui/react-icons';
 import clsx from 'clsx';
 import { useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


Kitenite (approved):
Nice this API makes things much easier! 
I will disagree with @drfarrell here. Instead of hiding, we should only disable. This is more expected behavior. Updated accordingly. Thank you!

PR #608, opened 2024-10-19
Tool-tip bug in Layer panel resolved

aPassie:
Tooltip errors in Layer Panel
Fixed Two errors :

1. The arrow pointing to the text loads in AFTER the rest of the tooltip
2 .Occasionally the tooltip will go too far to the right on hover. It's very inconsistent. Perhaps it's appending it to the end of the string location


Kitenite (changes requested):
This doesn't seem to address the issue. Could you confirm the issue is fixed?

Screenshot 2024-10-19 at 12 08 09 PM

PR #611, opened 2024-10-19
Fix maximize on windows and modify styles

abhiroopc84:
### Description

Modified the color of the close button to red on hover and active states, the shape of the buttons to fill the div, also the icon for maximize button. I also added in the unmaximize functionality to the maximize button, which was missing before. 

https://github.com/user-attachments/assets/0fc344ed-538f-472d-bd2c-b220b20a6ddb

https://github.com/user-attachments/assets/d02409f8-61f7-46be-9c46-c1b3d88e84cb

https://github.com/user-attachments/assets/b8e988e4-51b9-44fd-be12-23b8c4719c01

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/app/src/components/AppBar/WindowsControls.tsx
+++ b/app/src/components/AppBar/WindowsControls.tsx
@@ -5,7 +5,7 @@ import { MainChannels } from '/common/constants';
 import { WindowCommand } from '/common/models/project';
 
 export const WindowsControls = () => {
-    const [isMaximized, setIsMaximized] = useState(false);
+    const [isMaximized, setIsMaximized] = useState(true);
 
     if (process.platform !== 'win32') {
         return null;
--- a/app/src/components/AppBar/WindowsControls.tsx
+++ b/app/src/components/AppBar/WindowsControls.tsx
@@ -1,15 +1,20 @@
-import { Cross1Icon, MinusIcon, SquareIcon } from '@radix-ui/react-icons';
+import { useState } from 'react';
 import { Button } from '../ui/button';
 import { MainChannels } from '/common/constants';
 import { WindowCommand } from '/common/models/project';
 
 export const WindowsControls = () => {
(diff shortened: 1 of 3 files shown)


app/src/components/AppBar/WindowsControls.tsx
```diff
@@ -1,39 +1,47 @@
...
                 aria-label="Maximize"
             >
-                <SquareIcon className="h-3 w-3" />
+                <CopyIcon className="h-3 w-3 scale-x-[-1]" />
             </Button>
             <Button
                 onClick={() => sendCommand(WindowCommand.CLOSE)}
                 variant={'ghost'}
-                className="hover:bg-background-onlook/30 hover:text-foreground outline-border"
+                className="hover:bg-[#e81123] active:bg-[#e81123]/50 hover:text-foreground outline-border w-full h-full rounded-none"
```
Kitenite, line 44: Should we consider using our tokens here or is this the standard hex color @abhiroopc84?
Particularly, we have bg-red-500 as well.

https://github.com/onlook-dev/onlook/blob/e8e06499deb4e37967f33eed730cf7a6250a7bf7/app/src/index.css#L6

abhiroopc84: It's a standard hex color that I found across some major windows electron apps including vscode, spotify. We could use our tokens as well, it's just a personal choice I suppose. Let me know if you want me to use our tokens.

Kitenite: I see, let's keep them as is. Since these are perceived as system-level control, I don't want them to be affected by us refactoring our color tokens for example. 
Thanks for the thoughtful response.

Kitenite (approved):
Works great in testing! Thanks for implementing this :)

PR #614, opened 2024-10-19
Fix dom not ready crashing the app

abhiroopc84:
### Description

Fix an issue where the app was crashing when an element tries to call a method that requires the `dom-ready` event to be emitted before.  

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (commented):
I don't think this would be the case. The reference itself is non-null. The instance it references may be null. But it seems more of an issue due to the electron webview instance not loading a DOM yet. Has https://github.com/onlook-dev/onlook/pull/612 failed to fix the issue?

abhiroopc84:
#612 did fix the issue, but I just thought adding two additional try catch blocks just to catch those errors would be redundant when we can just have a state `domReady` that is a boolean value which we'll use to either pass the webviewRef when dom has loaded, or null when dom hasn't loaded yet, to the BrowserControl component. `domReady` state can be set to true when we run the existing `handleDomReady` function. This way we'll be fixing the issue in the same component where the webviewRef is defined originally, so we can always use the same fix even if we face a similar error in some other component instead of adding try catch blocks in each such error component.

app/src/routes/editor/WebviewArea/Frame.tsx
```diff
@@ -119,7 +121,7 @@ const Frame = observer(
         return (
             <div className="flex flex-col space-y-4">
                 <BrowserControls
-                    webviewRef={webviewRef}
+                    webviewRef={domReady ? webviewRef : null}
```
Kitenite, line 124: Ah I see what you did. Apologies, I missed this. This makes sense and adds a level of security. I'd hate to have an error thrown needlessly.

Kitenite (approved):
I like this. It adds just a bit of complexity but it's better than my duct tape solution. Well done!

Issue #566, opened 2024-10-16
[feat] Allow the ability to drag-in new elements from the toolbar


drfarrell:
<img width="302" alt="image" src="https://github.com/user-attachments/assets/4e3fbaf5-32ed-4e4c-80de-2d0adb1556cf">

Allow users to drag-in a standard size 100px X 100px Div or Text box from the bottom toolbar.

iNerdStack:
Hello @drfarrell, 
I am interested in working on this, can you give more context to this task and more specifics on this ?

alexandreantunesmnds:
Hi can i solve this issue ?

Kitenite:
@iNerdStack yes exactly

drfarrell:
> @iNerdStack yes exactly

To clarify, The user can click and drag the div button and it will drag a **div** of standard size onto the page, not a **div icon**. 

The same for the text box.

iNerdStack:
Understood. On it now

iNerdStack:
I have some updates while working on this issue

I am experiencing difficulties implementing drag and drop functionality on Electron's  [`<webview>`](https://www.electronjs.org/docs/latest/api/webview-tag) element. The toolbar draggable item works but doesn't fire an event when it is dragged over the `<webview>`.

## Current situation
- The [`<webview>`](https://www.electronjs.org/docs/latest/api/webview-tag) element does not appear to be a valid drop target. [https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API](url)
- Drag events (dragover, dragenter, drop) are not triggering on the webview as it is not part of the element.
- Attempts to use an overlay div on top of the webview to capture drag events  was also not unsuccessful.

```jsx
const Frame = () => {
  const handleDragOver = (e) => {
    e.preventDefault();
    console.log('Drag over');
  };

  const handleDrop = (e) => {
    e.preventDefault();
    console.log('Drop event fired');
    // Logic to insert element into webview
  };

  return (
    <div onDragOver={handleDragOver} onDrop={handleDrop}>
      <webview
        src="http://example.com"
        style={{ width: '800px', height: '600px' }}
      />
    </div>
  );
};
```

## What I've Tried
1. Adding drag event listeners directly to the [`<webview>`](https://www.electronjs.org/docs/latest/api/webview-tag) element
2. Creating an overlay div on top of the [`<webview>`](https://www.electronjs.org/docs/latest/api/webview-tag) to capture drag events
[...]

PR #627, opened 2024-10-22
feat: Allow the ability to drag-in new elements from the toolbar

iNerdStack:
### Description
Allow users to drag-in a standard size 100px X 100px Div or Text box from the bottom toolbar.

To enable users drag a standard box or text into the webview editor

- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

Related issue: [#566](https://github.com/onlook-dev/onlook/issues/566)


iNerdStack:
@drfarrell @Kitenite 

The draggable text and div box has been implemented and works in the [app/src/routes/editor/Toolbar/index.tsx](https://github.com/onlook-dev/onlook/pull/627/files#diff-f1c3650a2f6d8c20249858eb27e4d5f8d9aba0a1f0629a72f05714b08cf013c6) but needs some review & refactor. 

1. Where to place the standard elements for better code structure
2. Also a review on the insertion process 

``` javascript
    // Create and run insert action
        const insertAction: InsertElementAction = {
            type: 'insert-element',
            targets: [{
                webviewId: webview.id,
                selector: uuid,
                uuid
            }],
            element,
            location,
            editText: isTextElement
        };

        editorEngine.action.run(insertAction);

        setTimeout(async () => {
            // Force webview reload to get fresh state from code and editor
            webview.reload();

        }, 1500); // wait for code write to complete
```
This worked but without forcing the webview to reload after a few seconds delay after insertion, it seems to show 2 elements added to the editor and on the webview while the code is added once. If there's a better way I can update the PR to fix this.

Kitenite:
@iNerdStack , if the element is inserted with the data-onlook-inserted attribute, it will be removed when the code updates. here's an example

https://github.com/ashutoshdas-dev/onlook/blob/c0f81f472514085f3bd5895e54fd5f6df2ba2598/app/electron/preload/webview/elements/dom/insert.ts#L81

Here's the process that removes it. It's essentially deduplicated based on a key that's injected into code
https://github.com/ashutoshdas-dev/onlook/blob/c0f81f472514085f3bd5895e54fd5f6df2ba2598/app/electron/preload/webview/elements/dom/insert.ts#L144

Does this help?

iNerdStack:
https://github.com/user-attachments/assets/bc29141b-9f0f-4ca8-8dc7-3502fa07d573

@Kitenite Oh it did, I just realized it might be a bug not related to the method or changes in this PR,

In the video, I drew a box, and after a few seconds, you see about two of them created instead of one but not the case in the code pushed to the dom, it has just one div added, refreshing the webview seems to fix this temporarily.
  
It's the same issue I mentioned earlier. This issue is from the insertion and I believe if it can be fixed then the issue I am facing with drag and drop duplicate items will be solved too and the feature can be reviewed and merged.

This issue is also present on the published application and tested it on Mac and Windows version.

Kitenite:
Oh interesting. I could not reproduce the bug on my production instance (version 0.0.47) but I agree that the bug should be addressed. 
I'm nervous about reloading the page as a solution because you can lose the state of the webpage. If the page has load animation for example or some kind of open state, you would lose it. 

Do you have any suggestion for fixing this? Ideally, we would identify the elements coming in from code change and resolve it.

iNerdStack:
I also didn't like the idea of the webview reloading, i was hoping it would be a clue to how to solve the bug issue.

I will remove the webview reload method, that was just meant to temporarily fix the double-element issue.

On more context on how to replicate this bug.

1. Create a new project (possibly to test)
2. When all is done and loaded, ensure no layer is selected or select anywhere outside the webview canvas to ensure all selection made before is gone.
3.  Select the Insert div toolbar item, then draw a box anywhere outside any content or div box already displayed in the webview, any empty space would do.
4. After about 2 seconds you should see 2 items created.

If you still cant replicate, pull my branch/code to your local also and try and to drag and drop a box into the webview to see the issue. 

Also you can try and comment out this so the webview.reload() doesn't clear the issue.
 
``` javascript
     setTimeout(async () => {
            // Force webview reload to get fresh state from code and editor
            webview.reload();

        }, 1500); // wait for code write to complete
        
```

iNerdStack:
@Kitenite I have removed the webview reload logic and also fixed conflicts from new changes.

Please help test the implementation if you don't experience the double element issue, then its resolved

Kitenite (commented):
This is working really well already. A few things:

1. We have an existing insertElement functionality already which can be invoked with `editorEngine.insert.insertElement`. Is it possible to reuse this logic?
2. While dragging, we stop highlighting the elements being hovered. Is there a way to retain this behavior such that a user can see which element they are dragging the new element inside? 
3. Bonus: We create a stub element that shows up during drag to reorder to simulate where the new element would be. Is it possible to do the same for this? To insert a stub of the element inside such that the user can see what the end state would look like

https://github.com/inerdstack/onlook/blob/2bfbb098c03db22f4bea2da46c41ab4c94ff72f9/app/src/lib/editor/engine/insert/index.ts#L106

https://github.com/inerdstack/onlook/blob/2bfbb098c03db22f4bea2da46c41ab4c94ff72f9/app/src/lib/editor/engine/move/index.ts#L31

iNerdStack:
Okay, Let me work on that

iNerdStack:
https://github.com/user-attachments/assets/e7eb0584-a55e-4ebc-b0b7-4376d99ba2b7

Hello @Kitenite,
Some update.
1. <code>editorEngine.insert.insertElement</code> cannot be reused for our case unless refactored to allow specifying some extra properties in the createAction process and also createAction function allowing extra element properties object.

https://github.com/iNerdStack/onlook/blob/feat-draggable-toolbar/apps/studio/src/lib/editor/engine/insert/index.ts#L135
https://github.com/iNerdStack/onlook/blob/feat-draggable-toolbar/apps/studio/src/lib/editor/engine/insert/index.ts#L143

I created a new function called insertDroppedElement to handle insertion in our case also not to break existing insertElement function.

https://github.com/iNerdStack/onlook/blob/feat-draggable-toolbar/apps/studio/src/lib/editor/engine/insert/index.ts#L201

2. On hover highlight has been implemented and works with the gesturescreen mouseevents now.
3. I am still working on the stub element to show where the user would drop the element, stub already seems to be working when element is selected and moved but have not found that implementation, still looking into it.

I would appreciate a reference or idea of how the stub is already been used.

Kitenite:
1. Gotcha, it would be out of scope for you to try to refactor the function. I appreciate you giving this some consideration. 
2. Nice that works a lot better already
3. It would need to be somewhat custom to create a stub 
This is where it's currently being created. This logic would be different ofc. The stub would take the dimensions of the element and the computed insert location. Then on drop, we'd remove it. It can be out of scope for this PR. 

https://github.com/onlook-dev/onlook/blob/971d2de29612349aecfb6fcef06caff176c7ab43/apps/studio/electron/preload/webview/elements/move/drag.ts#L7

Kitenite (approved):
Amazing! This is a complex issue and you got it working very well. Clean implementation as well.

PR #632, opened 2024-10-23
icons: SVG to FC

### LINKED ISSUE #628: [feat] SVG and Icons organization
#### Describe the feature
Organizing SVGs and icons in a class for managing the bundle size. also, it is useful for maintenance.

hitaspdotnet:
### Description

Refactored SVG Imports

close #628 

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [X] Refactor
- [ ] Release
- [ ] Other


app/common/ide.ts
```diff
@@ -1,6 +1,5 @@
-import CursorIcon from '../src/assets/cursor.svg';
-import VsCodeIcon from '../src/assets/vscode.svg';
-import ZedIcon from '../src/assets/zed.svg';
+import { FC } from 'react';
+import { Icons, IconProps } from '@/components/icons';
```
Kitenite, line 2: This is a circular import. `common` should not have dependencies because it is used in the main electron process and the webview process. It was ok previously because we were importing svg but this is now pulling in more than that.

One solution to this is to move IDE into components and only keep the model in common. This is because the IDE type is used in the electron main process. Thoughts @hitaspdotnet

Kitenite: IDE is used in main here. We'd have to separate the Icons from the usage here. 

https://github.com/onlook-dev/onlook/blob/1794cea8bd7d9546abd0384c8132188b2376aceb/app/electron/main/code/index.ts#L78

Kitenite (changes requested):
Does this fail to build for you?

hitaspdotnet:
fixed in #641

Issue #599, opened 2024-10-18
[feat] Improve the project carousel design


drfarrell:
The original mockup had a number of features:

![image](https://github.com/user-attachments/assets/2e4ca98b-6882-4f4a-88eb-6eb843ba15cd)

1. The projects that are out-of-focus are smaller and have 60% opacity applied
2. You can see a portion of the top / bottom of the next / previous project in the carousel. The trickiness is that the previews are often different dimensions than another, so you may need to set a fixed gap in the carousel. 

Also, as we look to experiment adding in an acrylic background, it'd be good to fix the hover state overlay on the project card so it doesn't exceed the height of the project. 
<img width="805" alt="image" src="https://github.com/user-attachments/assets/99e8fdf4-7a10-42c2-9412-236b898c50e7">

abhiroopc84:
Can I work on this @drfarrell ?

sairiteshdomakuntla:
hello @drfarrell , can you assign this to me?

Kitenite:
Go for it @abhiroopc84!

Kitenite:
@sairiteshdomakuntla I'll give Abhiroop preference since he was first but are there any other I can assign you instead?

drfarrell:
@sairiteshdomakuntla do you want to take this one on? https://github.com/onlook-dev/onlook/issues/475 #475

sairiteshdomakuntla:
> @sairiteshdomakuntla do you want to take this one on? #475

yes, i think i can work on that

PR #643, opened 2024-10-24
Project carousel design improvement

abhiroopc84:
### Description

Improves the project carousel design in the following ways:
1. Projects that are out-of-focus are smaller and have 60% opacity applied
2. You can see a portion of the top / bottom of the next / previous project in the carousel

There is an issue with the slides starting index 2 having a style of `scale(0)` when scrolling initially. Also, I've tried adjusting the width and height to be unique for each project, but I wasn't able to pull it off, and I also wonder if the unique heights and widths of each project would give a uniform look. 

#599 

image

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite:
Hope you don't mind me pushing to your branch here. There seems to have been some extra layers added which caused the aspect ratio issue. You can see it fixed here. It's actually working except for one last thing, the images can be too short to show on the preview which makes sense. One last change is to align the image to the top or bottom if it's not in focused.


https://github.com/user-attachments/assets/cf7d1bc7-e14c-43b2-8794-99b06b83cc7c

Kitenite:
@abhiroopc84  Updated that here. 2 last small issue:
1. There's a flicker that happens because the alignment changes right away. It would be nice if we can delay that until after the animation is finished. Perhaps keep a "visualCurrentIndex" vs the currentIndex that only updates after the animation?
2. There's seems to be a new issue with scrolling quickly causing multiple scrolls. I think this did not exist before. Any idea how we can address this? 

Overall nice work! This feels like a more premium experience already. Just gotta make it smooth :)
https://github.com/user-attachments/assets/0059cd0c-bb9e-4286-8ed0-eb18a98a8d8f

Issue #651, opened 2024-10-25
[bug] Properties panel close animation is slightly buggy


drfarrell:
The properties panel elements trail behind in the animation

Some frames: 
<img width="451" alt="Screenshot 2024-10-25 at 2 14 00 PM" src="https://github.com/user-attachments/assets/986e1d59-f078-4848-b8d1-e4deef5a47ad">
<img width="438" alt="Screenshot 2024-10-25 at 2 14 33 PM" src="https://github.com/user-attachments/assets/b55abefa-4eda-4737-89a0-115156b8494c">
<img width="390" alt="image" src="https://github.com/user-attachments/assets/63fac042-bcc1-4cb3-9ba0-8fa2c44a4fd2">

Perhaps we just need to set overflow to hidden for the parent container?

PrakharJain1509:
@drfarrell Is this issue still open to work?

PrakharJain1509:
@drfarrell If its still open, Can I contribute in this?

PrakharJain1509:
@Kitenite @drfarrell Can you help me to just locate where is the code for this styles button. It would be really helpful, Thanks

PrakharJain1509:
@drfarrell @Kitenite I have created a PR resolving this. #654. Please review it and let me know if any changes are required.

PR #654, opened 2024-10-25
Debugged Properties panel close Animation Bug

PrakharJain1509:
# Fixing Overflow Issue in EditPanel Component

## Problem
While opening or closing the `EditPanel` tab, the content inside was overflowing beyond the boundaries of the panel. This caused the UI to break during animations.

## Solution
We applied **`overflow-hidden`** to critical containers and added **width/height constraints** to prevent content from escaping during transitions. We also ensured smooth transitions with CSS adjustments.

### Changes Made

1. **Apply `overflow-hidden` to the Main Wrapper:**
   - Ensures that the content stays within the panel boundaries during open/close transitions.
   ```jsx
   <div
       className={clsx(
           'fixed right-0 transition-width duration-300 opacity-100 bg-background/80 rounded-tl-xl overflow-hidden',
           editorEngine.mode === EditorMode.INTERACT ? 'hidden' : 'visible',
           !isOpen && 'w-12 h-12 rounded-l-xl cursor-pointer',
           isOpen && 'h-[calc(100vh-5rem)]',
           isOpen && selectedTab == TabValue.STYLES && 'w-60',
           isOpen && selectedTab == TabValue.CHAT && 'w-[22rem]',
       )}
   >

### DIFF
--- a/app/src/routes/editor/EditPanel/StylesTab/index.tsx
+++ b/app/src/routes/editor/EditPanel/StylesTab/index.tsx
@@ -104,7 +104,7 @@ const ManualTab = observer(() => {
                 <AccordionTrigger>
                     <h2 className="text-xs font-semibold">{groupKey}</h2>
                 </AccordionTrigger>
-                <AccordionContent>
+                <AccordionContent className="overflow-hidden">
                     {groupKey === StyleGroupKey.Text && <TagDetails />}
                     {renderGroupValues(baseElementStyles)}
                 </AccordionContent>
@@ -128,7 +128,7 @@ const ManualTab = observer(() => {
     return (
         editorEngine.elements.selected.length > 0 && (
             <Accordion
-                className="px-4"
+                className="px-4 overflow-hidden"
                 type="multiple"
                 defaultValue={[...Object.values(StyleGroupKey), TAILWIND_KEY]}
             >
--- a/app/src/routes/editor/EditPanel/StylesTab/index.tsx
+++ b/app/src/routes/editor/EditPanel/StylesTab/index.tsx
@@ -104,7 +104,7 @@ const ManualTab = observer(() => {
                 <AccordionTrigger>
                     <h2 className="text-xs font-semibold">{groupKey}</h2>
                 </AccordionTrigger>
-                <AccordionContent className="overflow-auto max-h-[50vh]">
+                <AccordionContent>
                     {groupKey === StyleGroupKey.Text && <TagDetails />}
                     {renderGroupValues(baseElementStyles)}
                 </AccordionContent>
@@ -128,7 +128,7 @@ const ManualTab = observer(() => {
     return (
         editorEngine.elements.selected.length > 0 && (
             <Accordion
-                className="px-4 overflow-hidden"
+                className="px-4"
                 type="multiple"
                 defaultValue={[...Object.values(StyleGroupKey), TAILWIND_KEY]}
             >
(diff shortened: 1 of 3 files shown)


PrakharJain1509:
@drfarrell @Kitenite here is the PR for that same, This has successfully debugged that issue and the text is now not coming out of the transition box. Please review it and let me know if any changes are required.

PrakharJain1509:
This was for #651 .

Kitenite (changes requested):
This is certainly better! However, I'm seeing some regression from this change:

1. the tailwind class box is larger than it should be. 
2. The editor panel no longer scrolls

Screenshot 2024-10-25 at 4 16 34 PM

PrakharJain1509:
@Kitenite Please check now, Scroll is working now, and the tailwind class is also in size.

PrakharJain1509:
@Kitenite Any updates sir?

app/src/routes/editor/EditPanel/StylesTab/index.tsx
```diff
@@ -104,7 +104,7 @@ const ManualTab = observer(() => {
                 <AccordionTrigger>
                     <h2 className="text-xs font-semibold">{groupKey}</h2>
                 </AccordionTrigger>
-                <AccordionContent>
+                <AccordionContent className="overflow-auto max-h-[50vh]">
```
Kitenite, line 107: Can you explain why this is necessary? Seems like we just need `overflow-hidden` on the root panel div?

app/src/routes/editor/EditPanel/StylesTab/index.tsx
```diff
@@ -128,7 +128,7 @@ const ManualTab = observer(() => {
     return (
         editorEngine.elements.selected.length > 0 && (
             <Accordion
-                className="px-4"
+                className="px-4 overflow-hidden"
```
Kitenite, line 131: Same with this

app/src/routes/editor/EditPanel/index.tsx
```diff
@@ -77,38 +77,38 @@ const EditPanel = observer(() => {
...
+                                <Icons.CrossS/>
                             </Button>
                         </div>
                     )}
                 </TabsList>
-                <Separator />
+                <Separator/>
                 <div className="h-[calc(100vh-7.75rem)] overflow-auto">
-                    <TabsContent value={TabValue.STYLES}>
+                    <TabsContent value={TabValue.STYLES} className="min-w-0 max-w-full min-h-0 max-h-full">
```
Kitenite, line 102: And these 2

Kitenite (commented):
Just one question, seems like there are more overflow-hidden implemented in code than what was suggested in the description. Is this correct? 
Also should we do the same thing in the LayersPanel?

Kitenite:
> @Kitenite Any updates sir?

Sorry I was working on a different PR. Left some comments but thanks for the quick turnaround!

app/src/routes/editor/EditPanel/StylesTab/index.tsx
PrakharJain1509: It creates issues with the Scroll if I do just that.

PrakharJain1509:
> > @Kitenite Any updates sir?
> 
> Sorry I was working on a different PR. Left some comments but thanks for the quick turnaround!

So I was not sure where the code was at first so was experimenting, So is it fine now ?

PrakharJain1509:
@Kitenite Did it for the layers too.

Kitenite (approved):
Working well! Thanks for being very responsive on this

PrakharJain1509:
@Kitenite Can you please add the hacktoberfest-accepted label also, for it to get counted.

Kitenite:
Ofc!

PR #688, opened 2024-10-30
New Package: @onlook/types

### LINKED ISSUE #687: Refactor Types: Extract studio types to new package
We can avoid future circular dependency problem by adding new package @onlook/types to share common types to other packages.

hitaspdotnet:
### Description

A package for shared types for onlook packages
close #687 

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [X] Refactor
- [ ] Release
- [ ] Other


apps/studio/package.json
```diff
@@ -70,6 +70,7 @@
         "use-resize-observer": "^9.1.0"
     },
     "devDependencies": {
+        "@onlook/models": "*",
```
Kitenite, line 73: This seems to be a different pattern than how the UI package is used?

packages/types/src/adapters/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface Prop {
+  name: string;
+  type:
+    | PropType.StringType
+    | PropType.BooleanType
+    | PropType.EnumType<string[]>;
+}
+
+export interface ComponentDescriptor {
```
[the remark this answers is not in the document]
hitaspdotnet: yes, for future framework not just react

apps/studio/package.json
Kitenite: Nvm its devDep because it's not runtime. But I wonder how that's going to play out with Zod since it'll be passed as an object but also generates types

packages/types/src/design-tokens/index.ts
```diff
@@ -0,0 +1,12 @@
+// W3C design tokens https://design-tokens.github.io/community-group/format
+
+export interface ColorToken {
```
Kitenite, line 3: Renamed file?

hitaspdotnet: it's new type for tokens system

apps/studio/package.json
hitaspdotnet: fixed

packages/models/src/ide/index.ts
```diff
@@ -0,0 +1,5 @@
+export enum IdeType {
```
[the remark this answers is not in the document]
hitaspdotnet: we need this `IdeType` in other models in `@onlook/models`. now, the `IDE` has reference to this package for this type.

hitaspdotnet:
> Overall looking good, a few thoughts & questions
> 
> 1. Seems like there are a few new files that should be renames
> 2. I'm implementing some zod models for chat. It does present an external dependency. I think it's still fine to put in models?

@Kitenite 
1. Those files are new. I need them for DnD integration. we can separate PR if it's required.
2. +1 for zod for chat or other types. yes, you can put anything that related to the models.

PR #700, opened 2024-10-31
Added copy-run-project command UI and funcionality

drfarrell:
### Description

- Empty state on the main project page
- From the project navigation page
- From the project in-editor top-bar

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/lib/editor/engine/canvas/index.ts
```diff
@@ -61,8 +61,16 @@ export class CanvasManager {
         this.saveSettings();
     }
 
-    get frames() {
-        return this.webFrames;
+    get frames(): (FrameSettings & { folderPath: string })[] {
```
Kitenite: I would not have done it like this. There are better ways to get the folder path.

apps/studio/src/routes/editor/WebviewArea/Frame.tsx
```diff
@@ -24,6 +27,7 @@ const Frame = observer(
         const RETRY_TIMEOUT = 3000;
         const DOM_FAILED_DELAY = 3000;
         const editorEngine = useEditorEngine();
+        const projectManager = useProjectsManager().project;
```
Kitenite, line 30: @drfarrell fyi, this is the easier way to get the current running project. From there you can get the path

PR #714, opened 2024-11-01
refactored models: zod integration

hitaspdotnet:

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [X] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/lib/editor/engine/history/index.ts
```diff
@@ -36,6 +36,7 @@ function undoAction(action: Action): Action {
             return {
                 ...action,
                 type: 'insert-element',
+                element: action.element,
```
Kitenite, line 39: This seems redundant. action is already spread

packages/models/src/actions/code.ts
```diff
@@ -14,61 +11,90 @@ export enum CodeActionType {
     UNGROUP = 'ungroup',
 }
 
-interface BaseCodeAction {
-    type: CodeActionType;
-    location: ActionElementLocation;
-    uuid: string;
-}
+const BaseCodeActionSchema = z.object({
+    type: z.literal(CodeActionType.MOVE),
```
Kitenite, line 15: This doesn't seem right. This should be the enum instead of the literal string.

packages/models/src/actions/code.ts
```diff
@@ -14,61 +11,90 @@ export enum CodeActionType {
...
+    location: MoveActionLocationSchema,
```
Kitenite, line 16: Neither is this. Seems like this should be ActionElementLocationSchema

packages/models/src/element/index.ts
```diff
@@ -1,37 +1,45 @@
...
+    encodedTemplateNode: z.string().optional(),
+    uuid: z.string(),
+});
+
+export const ParentDomElementSchema = BaseDomElementSchema;
+
+export const DomElementSchema = BaseDomElementSchema.extend({
+    tagName: z.string(),
+    styles: z.record(z.string(), z.string()),
+    parent: ParentDomElementSchema,
```
Kitenite, line 15: ParentDomElementSchema.optional()

packages/models/src/actions/code.ts
Kitenite: Change this to ActionElementLocationSchema causes circular dependency. Annoying...

packages/models/src/actions/code.ts
```diff
@@ -14,61 +11,90 @@ export enum CodeActionType {
...
-export interface CodeMove extends BaseCodeAction {
-    selector: string;
-    type: CodeActionType.MOVE;
-    location: MoveActionLocation;
-    childTemplateNode: TemplateNode;
-}
+export const CodeMoveSchema = BaseCodeActionSchema.extend({
+    selector: z.string(),
+    type: z.literal(CodeActionType.MOVE),
+    location: ActionElementLocationSchema.extend({
```
Kitenite, line 29: Is there no way to preserve MoveActionLocation? 
This is an important type and used often.

Kitenite: Pushing a fix for this

Kitenite (approved):
This is working and running as expected. Is zod type a hard requirement moving forward?

Kitenite:
Going to do one full UI/integration test before merging but confirmed types are as they were before.

Issue #674, opened 2024-10-28
[feat] Duplicate the current window into a new Window


drfarrell:
Ability for a user to create a copy of the current window and it's sizing about 100px to the right of the current window.

- Maintain the same dimensions, position on the page, dimensions, and all attributes of the window. 
- The button to do this with will be a "plus-circle" icon and the tooltop will say "Duplicate Window". This "Plus" button will later be a drop-down menu of creating a "Linked Window" (all scrolls and actions from the parent and child windows will be rendered across both windows) or an "Unlinked Window" (a separate instance / playground of the same window that isn't connected)

<img width="491" alt="image" src="https://github.com/user-attachments/assets/1e8a1657-7734-4199-a71d-4b67845a6bd8">

This is related to #619 and #196

Kitenite:
Aye aye @abhiroopc84

PR #716, opened 2024-11-02
Duplicate Window

abhiroopc84:
### Description

Initial functionality for duplicating an existing window.
#674 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (commented):
This is working great. The linked work in in progress but I wonder if using linkedIds is the best way to do it if we consider multiple-way linking. One alternative is to store link relationships on the project level instead of at the frame level. What do you think?

packages/models/src/projects/index.ts
```diff
@@ -20,6 +20,8 @@ export interface FrameSettings {
     url: string;
     position: RectPosition;
     dimension: RectDimension;
+    linkedIds?: string[];
```
Kitenite, line 23: I changed these to optional for backwards compatibility, fyi

apps/studio/src/routes/editor/WebviewArea/BrowserControl.tsx
```diff
@@ -161,6 +209,74 @@ function BrowserControls({
         );
     };
 
+    function renderDuplicateButton() {
+        return (
```
Kitenite, line 213: Added a simplified button instead of dropdown, fyi

Kitenite (approved):
Absolutely cooked with this one!

PR #757, opened 2024-11-06
Pre-Scan for required git and node dependencies

drfarrell:
### Description

- Initial UI for scanning whether a user as git and node installed on their machine. 
- Style improvements for the sign-in page
- Animations to animate between sign-in and the scan.

I put a force requirements screen section in Signin/index.tsx just a heads up for testing.
image

**What is missing**
- The actual detection and enforcing it not just at sign in, but at any point with the app.
- The install buttons actually working (this could likely use a loading state or progress bar or status.
- Handling user permissions if we need them to give us permissions to make these installs.

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/routes/signin/index.tsx
```diff
@@ -14,98 +15,284 @@ enum SignInMethod {
     GOOGLE = 'google',
 }
 
+enum SignInStep {
```
Kitenite, line 18: This login screen only appears when you're not logged in so I'm gonna have to move everything you have here into a new route.

Issue #97, opened 2024-07-31
[FEAT] Dimension labels next to a selection


drfarrell:
Add a small label outside of the selection outline to indicate to the user what the dimensions of the selection are. 

- The label contents should reflect the labels in the property panel on the right side. 
- The label should have 2px padding on the top and bottom, and 4px padding on the left and right. The corner rounding is 2px. 
- The fill color is Onlook's custom bg-red-600, and the label color is text-red-100. You should be able to reference those numbers in the project. 

<img width="436" alt="Screenshot 2024-07-31 at 7 38 30 PM" src="https://github.com/user-attachments/assets/4f0d629f-a035-4f4c-8e08-da3cda60fb4d">
<img width="594" alt="Screenshot 2024-07-31 at 7 38 38 PM" src="https://github.com/user-attachments/assets/da6f0cd2-f089-4f4a-afe0-a15c84e06234">
<img width="445" alt="image" src="https://github.com/user-attachments/assets/85bb837e-fc49-44de-922e-72b38f757744">

Kitenite:
There's missing functionalities for the hug and fill so can reopen or move the rest into a new issue @drfarrell 
There's some complexity involved with getting the hug and fill.

drfarrell:
I'll reopen because the context is still all here but some visual edits + still would like the hug + fill

- [ ] The top of the label should not be rounded corners – the bottom of the labels look good, but the top of the label should be straight so that it blends in perfectly with the border

<img width="445" alt="image" src="https://github.com/user-attachments/assets/b7434c7c-2715-4b58-a7b5-86f36ff14709">

- [ ] The character between the dimensions should be a multiplication sign (×) not an x. 

- [ ] The dimensions should reflect the hug or fill on an object as described above.

smirki:
I can work on this!

Kitenite:
Go for it! @smirki

Kitenite:
@smirki , how are things going? Do you need any support on this?

drfarrell:
@smirki any updates here? If you're not working on it anymore we may open it up to other contributors.

drfarrell:
@smirki I'm going to unassign you for now to leave room for other contributors who may want to pick this up. Let me know if you're still working on it!

xming13:
@drfarrell 
> The top of the label should not be rounded corners – the bottom of the labels look good, but the top of the label should be straight so that it blends in perfectly with the border
[...]

PR #764, opened 2024-11-09
Fix/dimension label styling update

xming13:
### Description

This PR address two issues described in https://github.com/onlook-dev/onlook/issues/97#issuecomment-2323614687

- [X] The top of the label should not be rounded corners – the bottom of the labels look good, but the top of the label should be straight so that it blends in perfectly with the border
- [X] The character between the dimensions should be a multiplication sign (×) not an x.

Note: it does not address the third issue, ie. "The dimensions should reflect the hug or fill on an object as described above."

Before
image

After 
image

edge case, when the text container's width is smaller than the box's width
image

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/apps/studio/src/lib/editor/engine/overlay/rect.ts
+++ b/apps/studio/src/lib/editor/engine/overlay/rect.ts
@@ -300,13 +300,29 @@ export class ClickRect extends RectImpl {
         const rectX = (width - rectWidth) / 2;
         const rectY = height;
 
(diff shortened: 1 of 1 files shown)


Kitenite (approved):
Looks great! Thanks for fixing this

PR #779, opened 2024-11-13
Support Next 15

Kitenite:

- [x] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite:
We have to make a decision between supporting every version of Next because they're backwards incompatible or do something different. Leaning into doing something different. Stay tuned.

Kitenite:
Next is supported in a different way

PR #782, opened 2024-11-14
Migrate ID system

Kitenite:
### Description

This is a very big change. We will no longer be using plugins to inject the `data-onlook-id` attribute. Instead, they will be written into code at runtime and cleaned up after. 

This will make it so that 

1. Onlook is compatible with any React codebase, no babel or swc plugins necessary
2. The IDs will remain consistent, increasing reliability of  generated code
3. Projects can be ran directly inside Onlook, no more external terminal needed

Drawbacks:

1. While Onlook is running, the codebase may look messier due to the added IDs
2. Projects NEED to run through Onlook instead of the command line

- [X] New feature
- [ ] Documentation update
- [X] Bug fix
- [X] Refactor
- [ ] Release
- [ ] Other


apps/studio/electron/main/run/helpers.ts
```diff
@@ -0,0 +1,91 @@
+import { type GeneratorOptions } from '@babel/generator';
+import * as t from '@babel/types';
+import type { TemplateNode, TemplateTag } from '@onlook/models/element';
+import * as fs from 'fs';
+import { customAlphabet } from 'nanoid/non-secure';
+import * as nodePath from 'path';
+import { VALID_DATA_ATTR_CHARS } from '/common/helpers/ids';
+
+export const ALLOWED_EXTENSIONS = ['.jsx', '.tsx'];
+export const IGNORED_DIRECTORIES = ['node_modules', 'dist', 'build', '.next', '.git'];
```
Kitenite, line 10: Maybe need to use gitignore files as well

apps/studio/electron/main/run/setup.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export function createMappingFromContent(content: string, filename: string) {
+    const ast = parseJsxFile(content);
+    if (!ast) {
+        return null;
+    }
+    return createMapping(ast, filename);
+}
+
+function createMapping(ast: t.File, filename: string): Record<string, TemplateNode> | null {
```
Kitenite, line 67: @iNerdStack , fyi. This is where information about the code is being computed. We can extend TemplateNode in order to add data such as array item. See the componentStack which does something similar by keeping a stack of the scope name

apps/studio/electron/main/run/terminal.ts
```diff
@@ -0,0 +1,116 @@
+import { MainChannels } from '@onlook/models/constants';
+import * as pty from 'node-pty';
```
Kitenite, line 2: @abhiroopc84 this is the only place where node-pty is being used. We'd swap it out if necessary.

PR #785, opened 2024-11-14
Added input/button features to toolbar

caleb8315:

I have added the button / user input features to the toolbar. 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/lib/models.ts
```diff
@@ -10,4 +10,6 @@ export enum EditorMode {
     PAN = 'pan',
     INSERT_TEXT = 'insert-text',
     INSERT_DIV = 'insert-div',
+    INSERT_MAIN_BUTTON = 'insert-main-button',
```
Kitenite, line 13: Instead of adding a new mode for each new element type which does not scale. What are your thoughts on grouping it all  (besides insert text because that has custom behavior) under something like `INSERT_ELEMENT`, then within that mode, we have a "sub-mode" which signifies what type of element it is?

apps/studio/src/routes/editor/Toolbar/index.tsx
```diff
@@ -45,6 +45,20 @@ const TOOLBAR_ITEMS: {
         disabled: false,
         draggable: true,
     },
+    {
```
Kitenite, line 48: @drfarrell were we hoping to have a dropdown list instead of adding 2 new buttons? The recall the design being slightly different

Kitenite (commented):
This is looking great! Just one comment on scalability and one on design for Daniel :)

apps/studio/src/routes/editor/Toolbar/index.tsx
drfarrell: that's right – @caleb8315 you originally mocked this up in one of your first versions, but we were imagining the UI looking like this with the up caret next to the div. I made some slight UI adjustments to the original mockup

image

It was very much inspired by what Figma has.

Screenshot 2024-11-15 at 1 04 03 PM

apps/studio/src/lib/models.ts
caleb8315: Sounds like a good idea to me! I'll give it a go!

Kitenite:
Closing for inactivity

Issue #644, opened 2024-10-24
[bug] Handle deleting next to an array


Kitenite:
#### Describe the bug

Example:
```
<div className=“flex flex-wrap gap-2”>
              {availableTags.map((tag) => (
                <button
                  key={tag}
                  type=“button”
                  onClick={() => toggleTag(tag)}
                  className={`px-3 py-1 rounded-full text-sm font-medium transition-colors ${
                    selectedTags.includes(tag)
                      ? “bg-orange-500 text-white”
                      : “bg-gray-100 text-gray-700 hover:bg-gray-200"
                  }`}
                >
                  {tag}
                </button>
              ))}
              <div className=“w-[30px] h-[30px] bg-[#E3F3FF]“></div>
            </div>
```

Kitenite:
Solved

PR #786, opened 2024-11-15
Fixed deleting element around array/map expressions

iNerdStack:
1. Fixed issue with deleting elements around array/map expressions
2. Added flag to non-static elements to prevent delete action

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

Related: [#644](https://github.com/onlook-dev/onlook/issues/644)


apps/studio/electron/preload/webview/dom.ts
```diff
@@ -1,10 +1,62 @@
...
+        const groups = children.reduce(
+            (acc, child) => {
+                if (!isValidHtmlElement(child)) {
+                    return acc;
+                }
+
+                const signature = [
+                    child.tagName,
+                    child.className,
+                    child.getAttribute(EditorAttributes.DATA_ONLOOK_ID),
```
Kitenite, line 26: My understanding: We're making an assumption that if an element has the same `data-onlook-id` as its sibling, it would be part of an array? 

One edge case is if an element is an instance and it's used twice, these would have the same signature?

```tsx
export default function Page() {
    return (
        <div>
            <Instance />
            <Instance />
        </div>
    );
}

function Instance() {
    return <div>Instance</div>;
}
```

apps/studio/electron/preload/webview/dom.ts
```diff
@@ -1,10 +1,62 @@
...
+            },
+            {} as Record<string, Element[]>,
+        );
+
+        // Mark elements that is part of a map/iteration, dynamic
+        Object.values(groups).forEach((group) => {
+            if (group.length > 1) {
+                group.forEach((element) => {
+                    if (element instanceof HTMLElement) {
+                        element.setAttribute(EditorAttributes.DATA_ONLOOK_DYNAMIC, 'true');
```
Kitenite, line 48: I would think that if a `DATA_ONLOOK_DYNAMIC_TYPE ` exist at all, it would be dynamic. So this may be redundant.

apps/studio/electron/preload/webview/dom.ts
```diff
@@ -65,6 +117,11 @@ export function buildLayerTree(root: HTMLElement): LayerNode | null {
 function processNode(node: HTMLElement): LayerNode {
     getOrAssignUuid(node);
 
+    const isDynamic = node.hasAttribute(EditorAttributes.DATA_ONLOOK_DYNAMIC);
+    const dynamicType: any = isDynamic
```
Kitenite, line 121: It maybe be worth keeping the types in an enum somewhere instead of have it be anything. In case we want to react a certain way to a type.

apps/studio/electron/preload/webview/dom.ts
Kitenite: Thought this is really close! I wonder if there's something we can do on the plugin side to populate these attributes

Kitenite (commented):
Some comments on the signature. This is an amazing solution and I think just the flagging can solve a lot of headache!

apps/studio/electron/preload/webview/dom.ts
iNerdStack: True. I initially started with DATA_ONLOOK_DYNAMIC only, but then decided to go further by trying to detect the type. You are correct - I will remove DATA_ONLOOK_DYNAMIC since it's now redundant.

apps/studio/electron/preload/webview/dom.ts
iNerdStack: oh yeah, That skipped me. I will update that

iNerdStack:
@Kitenite No not yet, I have some pending local changes to move dynamic type logic and also alerts when attempting to delete element in an array etc, i will make the push soon and inform you

Kitenite:
@iNerdStack awesome, just wanted to make sure I'm not blocking. Take your time :)

iNerdStack:
@Kitenite 
The PR is tested and ready for review

apps/studio/electron/main/code/diff/helpers.ts
```diff
@@ -79,3 +81,27 @@ export const jsxFilter = (
 export function generateCode(ast: t.File, options: GeneratorOptions, codeBlock: string): string {
     return generate(ast, options, codeBlock).code;
 }
+
+export function getDynamicType(elementPath: NodePath<t.JSXElement>): DynamicType | null {
```
Kitenite, line 85: My understanding is that this traverse up the entire element tree, is this too much? I would think if the immediate parent is an array then we will add it, otherwise we don't. In this case, we don't want to delete a but do we delete b and c?

Perhaps once we run into another parent that is a jsxelement, we can return.

```jsx
<>{arr.map(_ =>
   <a>
     <b/>
     <c/>
   <a/>
)}</>
```

iNerdStack:
``` javascript

export function getDynamicType(elementPath: NodePath<t.JSXElement>): DynamicType | null {
    const parentPath = elementPath.parentPath;
    
    if (t.isJSXElement(parentPath.node)) {
        return null; // Stop if we hit another JSX element
    }


    if (parentPath && t.isCallExpression(parentPath.node) &&
        t.isMemberExpression(parentPath.node.callee) &&
        t.isIdentifier(parentPath.node.callee.property) &&
        parentPath.node.callee.property.name === 'map') {
        return 'array';
    }

    return null;
}
```

@Kitenite does this solve the concern with returning early if the parent is another jsx element?

iNerdStack: Never mind, it didn't work as expected. root element in the array were not flagged as dynamic

Kitenite (approved):
This is awesome! I added a stack to track only the parent's dynamic type. Tested with nested maps and it works really well too!

Issue #486, opened 2024-10-09
[feat] Allow the window size to be changed while in Interact Mode


drfarrell:
Essentially let these handles be usable while in Interact mode.

<img width="780" alt="Screenshot 2024-10-09 at 12 15 21 AM" src="https://github.com/user-attachments/assets/d054d565-f780-4dde-ad48-ea7bd60370ff">

sahil-1729:
I want to work on this issue, can you assign it to me?

sahil-1729:
I’ve fixed the issue, and it’s working on my local machine. I’m ready to add tests, could you point me in the right direction on where to add test in the repo? Should I add them in the test folder [here]( https://github.com/onlook-dev/onlook/blob/main/app/tests/styles.test.ts ), or is there a other folder I should follow?

Kitenite:
Hey @sahil-1729,
We don't have UI tests yet. Though we really need them. 
Feel free to open the PR. If you'd like to add UI tests, that would be a really good separate task!

Kitenite:
Also keeping in mind that if dragging quickly, the resize can be buggy because the inreact window catches mouse event. This needs to be handled as well :)

sahil-1729:
As in the video, while shrinking the window size, the window is lagging, this bug is the one you told to be handled as well right?

https://github.com/user-attachments/assets/0b54e832-cb03-4454-9b46-ea175230f3e0

Kitenite:
Hey @sahil-1729 , yes! It's because in interact mode, the screen is interactable which catches the mouse events which is used for mousemove here
https://github.com/onlook-dev/onlook/blob/7fdf799e23763bda996b7634fd0711119739c47f/app/src/routes/editor/WebviewArea/ResizeHandles.tsx#L49

Kitenite:
@sahil-1729 , how is it going, were you able to get around this?

sahil-1729:
Apologies for the delay in updating. I've been exploring various approaches to resolve the issue but haven’t been able to get it to work. Could you please guide me on where to get started ?

Kitenite:
One way I could think of is to enable the gesture screen when resizing. This way, the cursor cannot be caught by the webview. Lmk if that makes sense.

sahil-1729:
[...]

PR #790, opened 2024-11-16
Feat : allow window size change in interact mode

sahil-1729:
### Description

Allow the window size to be changed while in Interact Mode

issue #486 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


package.json
```diff
@@ -44,5 +44,8 @@
     },
     "devDependencies": {
         "husky": "^9.1.6"
+    },
```
Kitenite, line 47: This should not have been added

Kitenite (changes requested):

apps/studio/src/routes/editor/WebviewArea/GestureScreen.tsx
```diff
@@ -174,7 +174,7 @@ const GestureScreen = observer(({ webviewRef, setHovered }: GestureScreenProps)
             <div
                 className={cn(
                     'absolute inset-0 bg-transparent',
-                    editorEngine.mode === EditorMode.INTERACT ? 'hidden' : 'visible',
+                    editorEngine.mode === EditorMode.INTERACT ? 'visible' : 'visible',
```
Kitenite, line 177: This is not what I meant at all. We don't want to keep it always visible because then interact mode does not work. It should be hidden during interact unless resizing is happening.

sahil-1729: Apologies for earlier. I misunderstood the requirements regarding the visibility during interact mode. I've now updated the feature so that the element is hidden during interact mode unless resizing is occurring, as you originally intended.

https://github.com/user-attachments/assets/9a90133f-880b-4756-93df-e3feb9e13851

Kitenite (approved):
Works great, thanks for adding this!

PR #817, opened 2024-11-27
Fix double style change from entering and blurring

Kitenite:

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite:
This actually doesn't address the issue. The issue is how we write to the stylesheet. If margin is written, and margin-bottom is also written, it somehow gets the computed property incorrectly.

PR #840, opened 2024-12-03
Move away from vite import

Kitenite:

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite:
This doesn't work. If anything, we should be use MAIN_ to keep keys off the browserview

PR #849, opened 2024-12-05
Fix handling of dynamic className in tailwind input

cleobnvntra:
### Description

- Added logic to display a warning message when dynamic variables are used in className.
- Displays Tailwind classes if no dynamic variables are present in the template literal.
- Shows a button in the UI to navigate to the relevant templateNode source code when dynamic variables are used.
- Updated the button name to improve user clarity.

### With template literals
Now properly displays Tailwind Classes even if template literals were used, but no dynamic variables were used.

**Selected div for reference**:
```typescript
<div
  className={`w-full min-h-screen flex items-center justify-center bg-red-50 relative overflow-hidden flex-col`}
  data-oid=":gge2lo"
>
// ...
</div>
```

**Tailwind Classes textarea**:
image

### Template literals with dynamic variables

**Selected div for reference**:
```typescript
const someClassName = "text-center text-gray-900 relative z-10";

<div className={`m-8 p-8 ${someClassName}`} data-oid="adqx-me">
// ...
</div>
```

**Tailwind Classes textarea**:
image

As asked, clicking the `Go to source` button in the textarea brings the user to the relevant template node in the source code.

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

### DIFF
--- a/apps/studio/electron/main/code/classes.ts
+++ b/apps/studio/electron/main/code/classes.ts
@@ -35,5 +35,23 @@ function getNodeClasses(node: t.JSXElement): string[] {
         return classNameAttr.value.expression.value.split(/\s+/).filter(Boolean);
     }
 
+    if (
+        t.isJSXExpressionContainer(classNameAttr.value) &&
+        t.isTemplateLiteral(classNameAttr.value.expression)
+    ) {
+        const templateLiteral = classNameAttr.value.expression;
+
+        // Checks if all classes in the expression are static
+        const allStatic = templateLiteral.expressions.length === 0;
+
+        // Returns the classes if all are static, otherwise returns a message
+        if (allStatic) {
+            const quasis = templateLiteral.quasis.map((quasi) => quasi.value.raw.split(/\s+/));
+            return quasis.flat().filter(Boolean);
+        } else {
+            return ['Dynamic classes detected.'];
+        }
+    }
+
     return [];
 }
--- a/apps/studio/src/routes/editor/EditPanel/StylesTab/single/TailwindInput/index.tsx
+++ b/apps/studio/src/routes/editor/EditPanel/StylesTab/single/TailwindInput/index.tsx
@@ -211,6 +211,19 @@ const TailwindInput = observer(() => {
         textarea.style.height = `${textarea.scrollHeight + 20}px`;
     };
 
+    const navigateToTemplateNode = async (oid: string | null) => {
+        if (!oid) {
+            console.error('No templateNode ID provided for navigation.');
+            return;
+        }
+
+        try {
+            await window.api.invoke(MainChannels.VIEW_SOURCE_CODE, oid);
+        } catch (error) {
+            console.error('Error opening TemplateNode in IDE:', error);
+        }
+    };
+
     useEffect(() => {
         if (instanceRef.current) {
             adjustHeight(instanceRef.current);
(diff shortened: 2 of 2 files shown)


apps/studio/electron/main/code/classes.ts
```diff
@@ -35,5 +35,23 @@ function getNodeClasses(node: t.JSXElement): string[] {
...
+
+        // Checks if all classes in the expression are static
+        const allStatic = templateLiteral.expressions.length === 0;
+
+        // Returns the classes if all are static, otherwise returns a message
+        if (allStatic) {
+            const quasis = templateLiteral.quasis.map((quasi) => quasi.value.raw.split(/\s+/));
+            return quasis.flat().filter(Boolean);
+        } else {
+            return ['Dynamic classes detected.'];
```
Kitenite, line 52: Could we return a special type instead? Maybe an error type with a flag and message? 
This way we can use special logic to handle the type. Something like this with better naming? 

My thinking eventually we may want to return objects to be used. This way we can extend the return types and add custom logic.

```typescript

interface ClassesReturnValue {
type: 'classes',
value: string[]
}

interface ErrorReturnValue {
type: 'error',
reason: string
}

type ReturnValue = ClassesReturnValue | ErrorReturnValue
```

What do you think @cleobnvntra ?

apps/studio/electron/main/code/classes.ts
```diff
@@ -35,5 +35,23 @@ function getNodeClasses(node: t.JSXElement): string[] {
         return classNameAttr.value.expression.value.split(/\s+/).filter(Boolean);
     }
 
+    if (
```
Kitenite, line 38: We could also return an error straight from here since we're only handling classes. Everything else passes through.

Kitenite (changes requested):
This is great! You got the core of the application down at this point. Just a few suggested improvements. Lmk what you think :)

apps/studio/electron/main/code/classes.ts
cleobnvntra: Yeah I think that's a good idea. It is easily customizable and extensible with an interface rather than just returning a string. Easily doable too as it seems like the functions involved here is not used anywhere else other than in `TailwindInput/index.tsx`.

Can add the implementation to the PR later after double checking for return type conflicts, and a better interface name 😆

apps/studio/electron/main/code/classes.ts
cleobnvntra: I reformatted the code into:

```typescript
    if (
        t.isJSXExpressionContainer(classNameAttr.value) &&
        t.isTemplateLiteral(classNameAttr.value.expression)
    ) {
        const templateLiteral = classNameAttr.value.expression;

        // Immediately return error if dynamic classes are detected within the template literal
        if (templateLiteral.expressions.length > 0) {
            return {
                type: 'error',
                reason: 'Dynamic classes detected. Dynamic variables in the className prevent extraction of Tailwind classes.',
            };
        }

        // Extract and return static classes from the template literal if no dynamic classes are used
        const quasis = templateLiteral.quasis.map((quasi) => quasi.value.raw.split(/\s+/));
        return {
            type: 'classes',
            value: quasis.flat().filter(Boolean),
        };
    }
```

This allows the code to exit the block early if it is an error (dynamic classes are used).
I am unsure if this is exactly what you meant. Can you please clarify further if I misinterpreted it? :)

Kitenite: This works great, thanks!

Kitenite (approved):
Hey this is working great! Gonna be super helpful since this is a super common occurrence. 

I just did some refactoring and moved the error reason outside of the classes value object so it doesn't interfere with edit history. Nice work!

PR #859, opened 2024-12-08
Fix terminal on windows

abhiroopc84:
### Description
Fix the powershell terminal on Windows by adding arguments to allow Powershell to execute scripts without any restriction.

image

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/routes/editor/EditPanel/WindowSettings/index.tsx
```diff
@@ -0,0 +1,142 @@
+import { Button } from '@onlook/ui/button';
```
Kitenite, line 1: This may be a different PR's work? I think you're just intending to pass the Bypass?

abhiroopc84: Yes, my bad

Issue #809, opened 2024-11-23
[feat] Window settings in the sidebar


drfarrell:
Build the right sidebar for the window that shows up when you have a window selected, but no divs in the window selected. 
![image](https://github.com/user-attachments/assets/311568a2-90d8-4976-8c7a-421dc765d67a)

We don't need the "duplicate" and "delete" functionality just yet so you can just build that UI and hide it. I think we'll probably want to write up a separate PR.

abhiroopc84:
Up for grabs? @drfarrell

drfarrell:
you know it! I'll assign you

PR #880, opened 2024-12-11
Feat/window settings

abhiroopc84:
### Description
image

Firstly, I wasn't sure where to store all the state variables for these settings so I stored them in the components itself, if needed we can move them somewhere else, which I'm guessing would be the case.

Second, I'm not able to apply the text-smallPlus class to the width and height input elements, if possible take a look at that.

#809 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell:
I went through and cleaned up some of the UI, but great stuff! @Kitenite I'll leave it to you to hide the window feature until we build the functionality.

PR #886, opened 2024-12-11
Enable linux build

razzeee:
### Description

This starts building and hopefully shipping for linux

The directory path is based on https://github.com/tutao/tutanota/blob/1337c4650a69be845a7765c9ee6d6c1da096041f/buildSrc/electron-package-json-template.js#L167 and should allow us to build a multiplatform flatpak easily

- [ ] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [x] Release
- [ ] Other

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
                 os:
                     - macos-latest
                     - windows-latest
-                    # - ubuntu-latest
+                    - ubuntu-latest
 
         steps:
             - name: Checkout Code


razzeee:
Hrm, builds not running on merge request makes this a bit harder to validate - I guess I could just temporarily change that for this PR. Please advise.

Kitenite:
https://github.com/onlook-dev/onlook/releases/tag/v0.1.4-linux @razzeee

razzeee:
doesn't seem to have the result of the `dir` target? Could that be caused by the artifact name?

razzeee:
For sake of completness, the appimage, seems to work fine, but I seem to not be able to login, as it opens the link in firefox and even after allowing, I never end up getting back to the app

Kitenite:
> doesn't seem to have the result of the `dir` target? Could that be caused by the artifact name?

Doesn't seem to be a valid build target. Can you double check? 
https://www.electron.build/linux#target

razzeee:
it's in the link you shared, last item?

.github/workflows/build.yml
Kitenite, line 1: Commenting out `release: ${{ startsWith(github.ref, 'refs/tags/v') }}` ([source](https://github.com/onlook-dev/onlook/pull/886/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R82))

and adding 

```
branches:
            - 'main'
            - 'your_branch'
```
Will make this build the artifact.

razzeee:
I assumed, the `dir` type would be more helpful then the deb/appimage, but it turns out, that it's also not allowing for multiplatform builds.

So just enabling the linux builds and starting to ship releases would unblock me, when it comes to building a flatpak for x64 - I'll change the PR accordingly

razzeee:
> For sake of completness, the appimage, seems to work fine, but I seem to not be able to login, as it opens the link in firefox and even after allowing, I never end up getting back to the app

Still might be the real problem

razzeee:
What's the reason for closing this?

Kitenite:
@razzeee sorry just cleaning up. I didn't know you were still working on this. I mentioned in a separate thread but I'm not able to test for Linux atm.

razzeee:
> @razzeee sorry just cleaning up. I didn't know you were still working on this. I mentioned in a separate thread but I'm not able to test for Linux atm.

the build works fine and the produced appimage seems to work, but I can't get past the login. not sure what exactly happened, but I think it did not redirect correctly.
when I run it now (0.1.4), it tries to open an unknown suprabase url, which has firefox complain with this:
image

Issue #811, opened 2024-11-24
[feat] Simplified Window URL 


drfarrell:
Combined with #810, but...

1. Make the window controls consolidate when the screen size is small, and 
2. When a user double-clicks the url is when they can type the url. Make the url field the full width if it's too small for the user to see what they are typing.

<img width="532" alt="image" src="https://github.com/user-attachments/assets/679a30e8-4f19-4629-9b1f-bc56fcb4a9b6">

drfarrell:
It'd be good to make a minimum window width too of say 280px. I think 300 is the standard, but just to be a little generous, I think 280 would be good.

abhiroopc84:
> It'd be good to make a minimum window width too of say 280px. I think 300 is the standard, but just to be a little generous, I think 280 would be good.

Any minimum height? Or should I let the height be as is?

Kitenite:
This seems done

Issue #810, opened 2024-11-24
[feat] Simplified top-bar controls


drfarrell:
This is lightly related to #809 

The goal here is to simplify the top-bar to be closer to what Figma has, while creating unique top-bar configurations depending on the state of the project / window. For example...
- Is the project connected to Onlook?
- What mode is the user in? Design or Interact

### 1. Click-and-drag
The area of the top-bar should be click-and-draggable, removing the need for the "drag" icon on the top of the window that currently exists. 

![Export-1732411361891](https://github.com/user-attachments/assets/85c6f248-71f6-4496-a170-029fe2ea81cc)

### 2. Consolidate window controls to a dropdown
Many of the functions we currently have can be consolidated to a dropdown on the far right corner of the page while in Design mode. 

<img width="1512" alt="Annotations" src="https://github.com/user-attachments/assets/d0d21932-6f2f-46ef-8f3a-02592f16fa0f">

### 3. Different controls during Interact mode 

The top bar controls should appear in interact mode according to the directions outlined here: 
<img width="1512" alt="Multiple Windows" src="https://github.com/user-attachments/assets/40cd2949-b01a-4580-b888-d5e2e01a324e">

Kitenite:
@drfarrell this is done?

Kitenite:
This seems done. Pls reopen if not.

PR #892, opened 2024-12-13
Feat/simplified top bar controls

abhiroopc84:
### Description

image
image
image
image

https://github.com/user-attachments/assets/349c864c-61a0-4936-bdb5-35147c1e7077

#810 
#811 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


drfarrell (changes requested):
This makes the UI look so so clean I'm excited about this! It will pair really well with the window UI in the sidebar... I'm excited about it!

I took a pass through it, added a bunch of style improvements that improve the UX and overall-feel, but I couldn't tackle two issues. 
1. The Design / Interact mode white top bar seems out of sync. I can't find out how to fix it so the bar slides over the interact mode like before.
Screenshot 2024-12-14 at 12 46 42 AM

2. In interact mode there is some odd click-behavior when I click the top bar. Perhaps it is some of the edits I made, but when I click to edit things it flickers to gray for one click, and is generally a bit buggy. 

Not sure if you can solve either of these @abhiroopc84, but if not, we can bring-in the big guns of Kiet.

PR #906, opened 2024-12-15
Hosting Pt.1

drfarrell:
### Description

A modal with two states, an initial "create link" state and then a "Copy / Unpublish / Update" state

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


packages/hosting/src/api/zonke.ts
```diff
@@ -0,0 +1,44 @@
...
+    const { environmentId, buildOutputDirectory, message } = options;
+
+    return deployToPreviewEnvironment({
+        environmentId,
+        buildOutputDirectory,
+        message,
+        // Default to 60 seconds for upload link expiration
+        uploadLinkExpirationOverride: 60,
+    });
+}
```
ngobzin11, line 40: I had to add a `packageJsonPath` property to support Remix builds. The default build output for Remix does not include the required modules, so we need to copy the `package.json` path so dependencies can be installed during deployment. See this CLI implementation:

https://github.com/zonke-inc/sdk-ts/blob/main/packages/cli/src/commands.ts#L37-L43

PR #924, opened 2024-12-23
Window controls implementation v1

### LINKED ISSUE #901: [feat] Implementing Window Controls v1
Because we simplified the top bar, we still need a way for the user to choose preset window sizes from the app. https://github.com/onlook-dev/onlook/issues/811

This is a continuation of this issue: https://github.com/onlook-dev/onlook/issues/809 where we actually need the functionality hooked up.

- Device – Allow the user to select a preset device dimension

- Orientation – allow the user to flip the orientation of the width / height in just a click. Also, if the user manually makes the width longer than the height, detect this and change the icon to what is appropriate. 

- Width / Height – Allow the user to specify the width and height of the window in pixels. 
  - You can borrow one of the number input functions that allows the user to use SHIFT+UP/DOWN arrow keys to increment by units of 10, otherwise up / down arrows for increments of 1.
  - Do not let the user input a number that is less than the minimum width / height of the browser windows

- Responsive is something we can hide for now, but it will assign sm:  lg: xl: etc. classes for tailwind values that are written to code from that specific window size. – This will be for V2 because it involves editing how tailwind classes and such are written in code.

- Device settings changes the window setting for whether its light / dark / device mode.

abhiroopc84:
### Description

Implementation of window controls and other minor fixes
#901 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


packages/models/src/projects/index.ts
```diff
@@ -17,6 +19,10 @@ export const FrameSettingsSchema = z.object({
     dimension: RectDimensionSchema,
     linkedIds: z.array(z.string()).nullable(),
     duplicate: z.boolean().nullable(),
+    orientation: z.nativeEnum(Orientation),
```
Kitenite, line 22: These will have to all be nullable for backwards incompatibility

Kitenite: I'm adding that and fixing the merge conflict atm

PR #929, opened 2024-12-25
Bug: Fixed App Crashing on Changing to a Invalid Url

kartik-raj7:
### Description

Fixed App Crashing on Changing to a Invalid Url, wrapped the getCleanUrl function in a try catch block to avoid the code from breaking when user inputs invalid url, instead in that case it shows a predefined url.

- [ ] New feature
- [ ] Documentation update
- [X] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (approved):
Nice thanks for spotting and fixing this!

PR #966, opened 2024-12-31
More secure env variables

Kitenite:

- [X] New feature
- [ ] Documentation update
- [X] Bug fix
- [x] Refactor
- [ ] Release
- [ ] Other


Kitenite:
This approach doesn't bundle the env variables. Thought it's a good try. 
For electron, the best way to fully secure secrets is to serve the secrets through a service and have the electron app query them after an auth flow.

PR #1134, opened 2025-01-27
fix: Apply Tailwind class to multiple selected elements

Nithishvb:
### Description

This PR addresses an issue where users were unable to apply Tailwind CSS classes to multiple selected elements simultaneously. Previously, only the first selected element received the assigned class, requiring manual updates for each element.

- [ ] New feature
- [ ] Documentation update
- [x] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other


Kitenite (changes requested):
Hey @Nithishvb awesome you got this figured out with no help!
While this technically works, this overwrites the tailwind classes completely. Instead, we would need to compute the tailwind class diff for each element and only add the new class instead. 

For example:
Element1: className="foo"
Element2: className="bar"

Currently, selecting both and edtinging Element1 to className="foo foo"  will turn Element2 to className="foo foo" as well 
We instead want it to be className="bar foo"

Does this make sense?

Nithishvb:
Hey @Kitenite, I’ve made the requested changes to the PR. The className handling now prevents overwriting other elements, and it computes the diff correctly.

Kitenite (approved):
This works great! Thank you for the quick turnaround

Issue #564, opened 2024-10-16
[feat] Resize the Properties panel


drfarrell:
Allow the user to resize the right Properties panel to a maximum width no greater than 1/3 of the full app window.

This is pretty much exactly what #164 was.

AkshatSharma5:
/assign
Also, can you please clarify the issue with a screenshot or something similar please.

drfarrell:
Sure thing! Assigned it to you. 

Here you can see the resizing works on the left panel but not the right one. That's what this issue is for.
![Export-1729108072933](https://github.com/user-attachments/assets/f0dc78ae-f220-43b5-a1f0-324ff6513cc0)

drfarrell:
Hey @AkshatSharma5 how's it going? Need any help?

vijayksingh:
hey @drfarrell it doesn't look like @AkshatSharma5 is working on this, can i assign it to myself and work on this ?

Kitenite:
@vijayksingh , go for it!

vijayksingh:
@Kitenite have one question the implementation has been changed as mentioned in #164 
There is a new `ResizablePanel` component which has been introduced but this component has left side alignment hardcoded. 

I don't think this should be the responsibility of this component and would like to move the side as prop outside of this component. 

What do you say ? 
```jsx
  <div
            className={clsx('absolute top-10 left-0 w-60 min-w-60 max-w-96')}
            ref={panelRef}
            style={{ width: `${panelWidth}px` }}
            id="layer-tab-id"
        >
            {children}
            <div
                className="top-0 h-full w-2 absolute right-0 cursor-ew-resize z-50 flex items-center justify-center"
                onMouseDown={startResize}
            />
        </div>
```

Kitenite:
Hey @vijayksingh I agree, We can expose the side as a 
side: "left" | "right" property such that it can be reused for both panels. This was the intention for ResizablePanel to be.

drfarrell:
@vijayksingh how's it going? Need help with anything? We may have some other contributors who'd like to take it on, so just let me know if you plan on putting in a PR or if you've run into any issues.

kartik-raj7:
Hey Can I work on this issue

drfarrell:
Hey @kartik-raj7 – just had a user request this today. Are you still workin' on this? Or should we open it up to other people to contribute?

kartik-raj7:
[...]

PR #1146, opened 2025-01-28
feat: Resize properties panel

akshat-ilen:
### Description

This PR refactors the resizable panel to have the capability for both left and right panels. It also adds the capabilty to resize the properties panel on the left.
Issue - #564 

### Testing

https://github.com/user-attachments/assets/b55879ff-7afe-4d67-8c53-8ab681c2d73c

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [X] Refactor
- [ ] Release
- [ ] Other


apps/studio/src/routes/editor/EditPanel/index.tsx
```diff
@@ -103,16 +103,14 @@ const EditPanel = observer(() => {
         <div
             id="style-panel"
             className={cn(
-                'fixed right-0 transition-width duration-300 opacity-100 bg-background/80 rounded-tl-xl overflow-hidden',
+                'right-0 absolute transition-width duration-300 opacity-100 bg-background/80 rounded-tl-xl overflow-hidden',
                 editorEngine.mode === EditorMode.INTERACT ? 'hidden' : 'visible',
-                isOpen ? 'h-[calc(100vh-5rem)]' : 'w-10 h-10 rounded-l-xl cursor-pointer',
-                isOpen && selectedTab == EditorTabValue.STYLES && 'w-60',
-                isOpen && selectedTab == EditorTabValue.CHAT && 'w-[22rem]',
```
Kitenite, line 110: Slightly out of scope but is there a way we can still retain this behavior? 
When switching tab value, we still want to use these default widths, then user can override by resizing. One way I can think of is to move this to the top level. What are your thoughts @akshat-ilen?

Kitenite (commented):
This is very nice work! I love the refactor as well. One last comment let me know your thoughts :)

apps/studio/src/routes/editor/EditPanel/index.tsx
akshat-ilen: Instead, I would recommend to have resizablePanel in the editPanel. What are your thoughts?

Kitenite: I think that makes a lot more sense. Then we keep the right separation of concern for tab logic being inside the EditPanel. I like it!

Issue #1259, opened 2025-02-04
[feat] Add properties panel functionality


Kitenite:
#### Describe the feature

Continuation of https://github.com/onlook-dev/onlook/pull/1132
The properties panel UI has now been created. We will want to support adding properties

1. Support boolean type properties
2. Support text type properties
3. For complex properties, we can show an edit code button

This is similar to how the TailwindCSS box works. TailwindCSS we were reading the className property. This will read all of them similarly.

![Image](https://github.com/user-attachments/assets/cbafeef4-0227-4b52-b89b-1e719703ad6e)

Nithishvb:
@Kitenite, I’m on this!.

Kitenite:
💯

PR #421, opened 2024-09-29
[poc] IDE in canvas (codemirror)

Kitenite:
### What is the purpose of this pull request? 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

PR #420, opened 2024-09-29
[poc] IDE in canvas (monaco)

Kitenite:
### What is the purpose of this pull request? 

- [X] New feature
- [ ] Documentation update
- [ ] Bug fix
- [ ] Refactor
- [ ] Release
- [ ] Other

PR #1276, opened 2025-02-06
feat: Add properties panel

Nithishvb:
## Description

Closes: #1259

This PR introduces a properties panel that allows users to view and edit the properties of selected elements.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release

### DIFF
--- a/apps/studio/electron/main/code/diff/style.ts
+++ b/apps/studio/electron/main/code/diff/style.ts
@@ -38,3 +38,46 @@ function insertAttribute(element: t.JSXOpeningElement, attribute: string, classN
     const newClassNameAttr = t.jsxAttribute(t.jsxIdentifier(attribute), t.stringLiteral(className));
     element.attributes.push(newClassNameAttr);
 }
+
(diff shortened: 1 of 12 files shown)


drfarrell:
Notes: 
1. This would work really well with #421 or #420. So when the user hits "Edit code" they're able to bring it up. 
2. It seems like we'll want to implement this on the left panel, but it doesn't feel completely deserving of a full panel tab in of itself because there just aren't many properties for most objects. @Kitenite we should discuss the design / feel of where this should go a bit.
3. No functionality to add properties, but perhaps we make that a button that just sets up a prompt in the chat saying "Add a property to this component that does..."

Kitenite:
> 2. It seems like we'll want to implement this on the left panel, but it doesn't feel completely deserving of a full panel tab in of itself because there just aren't many properties for most objects. @Kitenite we should discuss the design / feel of where this should go a bit.

I think this could be deserving of a left tab depending on the element. But what about right tab or a subsection of style tab? This will be very useful for components + regular HTML tags

Issue #1241, opened 2025-02-04
[feat] Image management functionality


Kitenite:
#### Describe the feature

Continuation of https://github.com/onlook-dev/onlook/issues/1063
Currently, the UI is built and hidden on the LayersPanel side

1. The image tab should scan for the standard Next.js assets folder and find any images that can be used. 
2. The images should be loaded into the sidepanel
3. Allow the ability to add image which will put it in the expected next.js standard folder
4. Add a search bar that just filters based on filename

Optional (can be separate PR)
1. Renaming the image should rename it but also find usage instance and refactor those as well
2. Support dragging the image into the frame which should add it to the target div's background, similar to the fill effect in the styles Panel

![Image](https://github.com/user-attachments/assets/9e63a5a2-6970-42c7-9bb2-221d379c15ee)
![Image](https://github.com/user-attachments/assets/aa9547da-58da-4489-b2dd-6e2fe5c7c0ab)

Nithishvb:
Hey @Kitenite, can i take this one?

Kitenite:
@Nithishvb , @spartan-vutrannguyen  is picking this one up already. Would you want to try this instead?
https://github.com/onlook-dev/onlook/issues/1259

Nithishvb:
Hey @Kitenite, Yes, I’d like to take this  #1129  new one instead

Kitenite:
@Nithishvb sounds great could you leave a comment in the issue?

Kitenite:
Done!

PR #1278, opened 2025-02-06
feat: add assets management

spartan-vutrannguyen:
## Description

Currently, these features below are built and hidden on the LayersPanel side

- Scan for the standard Next.js assets folder and find any images that can be used and loaded images into the sidepanel
- Allowed the ability to add image which will put it in the expected next.js standard folder
- Added a search bar that just filters based on filename

## Related Issues

- Related to #1241
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Release

https://github.com/user-attachments/assets/b6adaf45-8b15-4fd8-9963-9230b50d0a7e

### DIFF
--- a/apps/studio/electron/main/assets/index.ts
+++ b/apps/studio/electron/main/assets/index.ts
@@ -21,10 +21,20 @@ async function scanImagesDirectory(projectRoot: string): Promise<ImageContentDat
                     const imagePath = path.join(imagesPath, entry.name);
                     const image = readFileSync(imagePath, { encoding: 'base64' });
 
+                    const mimeTypes: { [key: string]: string } = {
+                        '.jpg': 'image/jpeg',
+                        '.jpeg': 'image/jpeg',
+                        '.png': 'image/png',
+                        '.gif': 'image/gif',
+                        '.webp': 'image/webp',
+                        '.svg': 'image/svg+xml',
+                        '.ico': 'image/x-icon',
+                    };
+
                     images.push({
                         fileName: entry.name,
-                        content: `data:image/png;base64,${image}`,
-                        mimeType: 'image/png',
+                        content: `data:${mimeTypes[extension]};base64,${image}`,
+                        mimeType: mimeTypes[extension],
                     });
                 }
             }
@@ -46,7 +56,7 @@ export async function scanNextJsImages(projectRoot: string): Promise<ImageConten
     }
 }
 
-export async function uploadImage(
+export async function saveImageToProject(
     projectRoot: string,
     image: string,
     fileName: string,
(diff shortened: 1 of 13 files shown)


apps/studio/electron/main/assets/index.ts
```diff
@@ -0,0 +1,63 @@
...
+                // Common image extensions
+                if (
+                    ['.jpg', '.jpeg', '.png', '.gif', '.webp', '.svg', '.ico'].includes(extension)
+                ) {
+                    const imagePath = path.join(imagesPath, entry.name);
+                    const image = readFileSync(imagePath, { encoding: 'base64' });
+
+                    images.push({
+                        fileName: entry.name,
+                        content: `data:image/png;base64,${image}`,
```
Kitenite, line 26: This seems to only handle png encoding

apps/studio/electron/main/assets/index.ts
```diff
@@ -0,0 +1,63 @@
...
+export async function scanNextJsImages(projectRoot: string): Promise<ImageContentData[]> {
+    try {
+        return await scanImagesDirectory(projectRoot);
+    } catch (error) {
+        console.error('Error scanning images:', error);
+        throw error;
+    }
+}
+
+export async function uploadImage(
```
Kitenite, line 49: I would rename this to something like saveImageToProject upload implies a different action

apps/studio/src/lib/editor/engine/image/index.ts
```diff
@@ -1,11 +1,84 @@
...
+        if (!projectFolder) {
+            console.error('Failed to delete image, projectFolder not found');
+            return;
+        }
+        await invokeMainChannel<string, string>(
+            MainChannels.DELETE_IMAGE,
+            projectFolder,
+            imageName,
+        );
+        this.scanImages();
```
Kitenite, line 80: Could use some error handling here

drfarrell:
Great stuff @spartan-vutrannguyen! Here are some notes: 

1. Can we make the search faster? is there a faster way to sort through what's in there near instantly? 
2. Can we add the functionality? Rename, delete with a warning, etc.? Let me know if you're missing anything from us for direction there. 
3. If I open the image upload and then don't choose anything, I shouldn't get this warning
image
4. Let's add a tooltip over the plus button that says "Upload an image"
5. Can we add "Drag and drop" functionality where if the user has the image tab open, they can drop an asset in? Similar to how we do it in the Chat input in the lower right corner, but in this case, it actually puts the image in the folder. See how when the user drags the image, the background color of the input turns teal. 
6. Can we put the detail menu to the right of the three dots? 
Screenshot 2025-02-09 at 1 18 36 PM
age in the folder. 

Eventually we'll want people to be able to drag those images onto the page and insert them.

PR #1325, opened 2025-02-10
Fix: Hide delete options in main window

Nithishvb:
## Description

This PR addresses an issue where users could select the main window and then attempt to delete it, which would result in a no-op. This behavior was misleading and could lead users to believe the application was not functioning correctly.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release

Screenshot 2025-02-10 205529


Kitenite (commented):
Nice nice! This is a good reminder of how much dead logic we have as well. Thank you once again @Nithishvb!

Issue #1350, opened 2025-02-11
[bug] Only deletes the first element selected when multi-selecting elements to delete


thughey:
When selecting multiple elements to delete – it only deletes the first element selected.

Here's how I'm running into it:

https://github.com/user-attachments/assets/4b40e805-efa1-4ccb-8c61-91361a23aedb

It only deletes the first element selected when multi-selecting elements in both the layer panel & on the canvas.

Kitenite:
This is now solved thanks to @Nithishvb! https://github.com/onlook-dev/onlook/pull/1353

PR #1353, opened 2025-02-12
fix: Enable deletion of multiple selected elements

Nithishvb:
## Description

Closes: #1350

This PR addresses an issue where only a single selected element was deleted, even when multiple elements were selected. The `delete()` function has been updated to iterate through all selected elements, enabling bulk deletion.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release


Kitenite (commented):
Awesome work! Thanks for taking this up :)

PR #1355, opened 2025-02-13
fix: messed up layer tree when copy/group/ungroup nested element

### LINKED ISSUE #994: [bug] Copy-pasting sometimes screws up the layers panel
Copy-pasting divs with multiple elements nested within them can sometime cause a weird layer overlap on the layers panel
image

spartan-vutrannguyen:
## Description

- When user copy and paste nested elements, it makes the layer tree messed up sometimes
-  Root-cause:  Does not provide a fresh element for children in 
## Related Issues

- fixes #994 
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release

## Testing

https://github.com/user-attachments/assets/0ca4e216-ab66-4e64-a0a6-cadcb6f7072f

### DIFF
--- a/apps/studio/src/lib/editor/engine/ast/index.ts
+++ b/apps/studio/src/lib/editor/engine/ast/index.ts
@@ -8,6 +8,7 @@ import { LayersManager } from './layers';
 
 export class AstManager {
     private layersManager: LayersManager = new LayersManager();
+    private templateNodeCache = new Map<string, TemplateNode>();
 
     constructor(private editorEngine: EditorEngine) {
         makeAutoObservable(this);
@@ -67,7 +68,6 @@ export class AstManager {
             console.warn('Failed to processNodeForMap: No oid found');
             return;
         }
-
         const templateNode = await this.getTemplateNodeById(node.oid);
         if (!templateNode) {
             console.warn('Failed to processNodeForMap: Template node not found');
@@ -87,11 +87,11 @@ export class AstManager {
             return;
         }
 
-        if (templateNode.dynamicType) {
+        if (templateNode.dynamicType && !node.dynamicType) {
             node.dynamicType = templateNode.dynamicType;
         }
 
-        if (templateNode.coreElementType) {
+        if (templateNode.coreElementType && !node.coreElementType) {
             node.coreElementType = templateNode.coreElementType;
         }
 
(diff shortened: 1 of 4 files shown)


apps/studio/src/lib/editor/engine/group/index.ts
```diff
@@ -56,6 +56,14 @@ export class GroupManager {
             }
             return null;
         }
+        // If there is only one element, we can't group it
+        if (elements.length === 1) {
```
spartan-vutrannguyen, line 60: @Kitenite @drfarrell I added this validation because I think we should only allow grouping from 2 elements. Correct me if I was wrong.

Kitenite: Grouping 1 element is ok because you may want to wrap a div around something. But I like the initiative

apps/studio/src/lib/editor/engine/ast/index.ts
```diff
@@ -87,11 +87,11 @@ export class AstManager {
             return;
         }
 
-        if (templateNode.dynamicType) {
+        if (templateNode.dynamicType && !node.dynamicType) {
             node.dynamicType = templateNode.dynamicType;
         }
 
-        if (templateNode.coreElementType) {
+        if (templateNode.coreElementType && !node.coreElementType) {
```
Kitenite, line 94: @spartan-vutrannguyen what's the reasoning for this?

apps/studio/src/lib/editor/engine/ast/index.ts
```diff
@@ -8,6 +8,7 @@ import { LayersManager } from './layers';
 
 export class AstManager {
     private layersManager: LayersManager = new LayersManager();
+    private templateNodeCache = new Map<string, TemplateNode>();
```
Kitenite, line 11: This cache never gets cleared. Is this going to be a memory leak?

Kitenite: Since the invoke will call another cache in main process, gonna remove this.

apps/studio/src/lib/editor/engine/ast/index.ts
Kitenite: Gonna remove in case the node element type changes and we want to keep it updated

Issue #1365, opened 2025-02-14
[bug] Need to handle when port 3000 is taken better


Kitenite:
#### Describe the bug
The problem is that people aren’t realizing when their app is on a different port than the webview window. In this case they don’t find the terminal. We need to warn them when the port is taken or if we can discover the new port, give them an option to change their frame. 

Either pop up a warning modal with a button that will change the port or cancel

drfarrell:
Show a modal to the user telling them to "stop running their app on localhost:3000" when there is an app detected on localhost:3000. Perhaps make it persist while the port is still occupied. We could include a refresh button in the lower right corner of the modal to check.

Nithishvb:
@Kitenite, I’m on this!.

PR #1397, opened 2025-02-16
feat: Improve port conflict handling

Nithishvb:
## Description

Closes: #1365

This PR introduces a warning modal that appears when the application detects that port 3000 is already in use.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release


drfarrell:
@Kitenite – 
* Added styling to the modal
* Added a refresh button for people to check to see if the port is occupied still
* Added styling / functionality for the run button to change to "Port is in Use". The user also has to click that modal. 

One thing is the CTA of the open modal is to "change port" but I almost feel like we shouldn't do that because we don't automatically redirect the window to the new "changed" port. I'd prefer if we were the ones to own that port, but I'll leave it up to your recommendation on what we want to do with that. 

Ideally, we force people to stick with localhost:3000 but we'll need to handle if they change the port via the current CTA in the modal. If they do that, then we need to add functioanlity in that updates both the window and the project port number.

Kitenite (approved):
Awesome thanks for the great work as always @Nithishvb!  Sorry it took so long to review.

apps/studio/src/routes/editor/Toolbar/Terminal/RunButton.tsx
Kitenite, line 1: Note: I was considering pulling this into its own button that just swaps with the run button when port is not available but decided against it for now to take advantage of the nice styling we already had

PR #1423, opened 2025-02-18
Support hosting proxy

Kitenite:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release


Wh1isper:
Hi, Thanks for this awesome project!

I do want to selfhost onlook and use our own api key. I'd like to use llm api proxy which gives us an idea of how much each user is using, but I noticed that there doesn't seem to be a document for supabase proxy function, are there any plans to open source this part? Or any guide for us to deploy our own?

Issue #467, opened 2024-10-06
[feat] Multiple Language support - Japanese


drfarrell:
Translate the key elements and labels on the app into Japanese 

A bunch of users from Japan tried out our product. Arigatooooo

PR #1426, opened 2025-02-18
Japanese, Chinese Support + Changing Interact --> Preview

drfarrell:
## Description

* Set the foundation for some UI elements to be translated with the goal of starting with Japanese. 

## Related Issues

#467 

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Release

## Additional Notes

The goal is to setup the english user interface for translation to Japanese, starting with some core elements that are very unlikely to change soon.


apps/studio/src/i18n.ts
```diff
@@ -10,6 +11,9 @@ const resources = {
     ja: {
         translation: jaTranslation,
     },
+    zh: {
+        translation: zhTranslation,
```
Kitenite, line 15: Why is this called zh?

drfarrell: The ISO 2-letter language code for Chinese is zh (Zhong-guo)

ch is a different language code

apps/studio/src/routes/editor/EditPanel/ChatTab/ChatInput.tsx
```diff
@@ -293,11 +295,8 @@ export const ChatInput = observer(() => {
                     disabled={disabled}
                     placeholder={
                         disabled
-                            ? projectsManager.runner?.isRunning ||
```
Kitenite, line 296: Regression here

PR #1429, opened 2025-02-18
feat: position control

### LINKED ISSUE #545: [feat] Enable Absolute Positioning in the editor as a default
Make it so you can start with absolutely positioned elements and then gradually add in layout. We need to handle absolute positioning on the editor side panel as well. So, a reactive section based on layout type. 

image

This can be three PRs: 
1. The UI in the picture
2. The logic of the UI controls to work in the properties panel
3. The dragging logic to allow for elements to be positioned on the canvas (and have their position tied to the absolute

spartan-vutrannguyen:
## Description

- Allow setup position of an element
## Related Issues

- Closes #545
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Release


apps/studio/src/routes/editor/EditPanel/StylesTab/compound/PositionInput.tsx
```diff
@@ -0,0 +1,180 @@
...
+        );
+    };
+
+    const centerElement = async () => {
+        const selectedStyle = editorEngine.style.selectedStyle;
+        if (!selectedStyle) {
+            return;
+        }
+
+        const element = editorEngine.elements.selected[0];
```
Kitenite, line 57: This should probably just work for multiple elements by wrapping in a loop of all selected elements

apps/studio/src/routes/editor/EditPanel/StylesTab/compound/PositionInput.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+        const parent: DomElement | null = await webview.executeJavaScript(
+            `window.api?.getParentElement('${element.domId}')`,
+        );
+        if (!parent) {
+            return;
+        }
+
+        // Calculate center position
+        const centerX = (parent.rect.width - element.rect.width) / 2;
```
Kitenite, line 75: This should be a percentage instead so it's still actually responsive. Something like this
```
.child{
  position: absolute;
  top: 50%;
  left: 50%;
  transform: translate(-50%, -50%);
}
```
https://medium.com/front-end-weekly/absolute-centering-in-css-ea3a9d0ad72e

Kitenite: Actually, perhaps this is better since we don't have a transform editor

```
{
position: absolute;
left: 0;
right: 0;
margin: auto
}
```

Kitenite:
A couple small comments: 

1. The red bars should be clickable
2. Should round the values to the nearest pixel value

Next step:
3. When in absolute mode, we should be able to drag the element to change the absolute position instead of reordering 

Screenshot 2025-02-19 at 11 48 01 AM

drfarrell:
1. Like what kiet said, the values should snap to pixel values.
Screenshot 2025-02-19 at 12 53 07 PM

2. Sometimes it takes two attempts to change a value in the box for it to apply. 
3. clicking on the red bars should "activate" that mode and make the number input white (or text-foreground-active). If the red bar is not clicked, then the value should be ignored. It's not enough to change it to "auto" because it seems to default to "0px" which would be the same as explicitly defining that it should be at 0px from the side of the input. 

4. The individual bars in the middle should be clickable. See in the example image how if the user hovers over the middle bar, it highlights it as if it could be clicked, and if they hover over the horizontal bar the same happens. 

Screenshot 2025-02-19 at 1 03 32 PM

Screenshot 2025-02-19 at 1 03 12 PM

Meaning it's possible to have both selected (centering relative to the parent)
Screenshot 2025-02-19 at 1 04 06 PM

drfarrell:
Had a quick change to review this and going to drop some feedback. Generally, the dragging is working pretty well! I like it and can feel that this is going to be a really unique feature. 

Forgive me if some of this is stuff you are already working on!

1. The object should snap to the nearest pixel on dragging to prevent the 0.584498 sub pixels and such.
2. There seems to be a flicker when the object is dragged. The old position flickers for a second. 
3. Can we make the click-area around the red side buttons a bit larger? so the lines remain the same thickness, but have more clickable "padding" around them.
4. If I click the center of the textbox, it seems to jump back to the upper left corner.
5. The anchors (the red lines) seem to turn on when I drag an item around. They should really only be activated by the user to "lock" their relationship in place to the edges of the window.
6. This may be a bit tough, but if it's possible to update the numeric values in real-time, that would be awesome, but that could be a v2 feature.

spartan-vutrannguyen:
> Had a quick change to review this and going to drop some feedback. Generally, the dragging is working pretty well! I like it and can feel that this is going to be a really unique feature.
> 
> Forgive me if some of this is stuff you are already working on!
> 
> 1. The object should snap to the nearest pixel on dragging to prevent the 0.584498 sub pixels and such.
> 2. There seems to be a flicker when the object is dragged. The old position flickers for a second.
> 3. Can we make the click-area around the red side buttons a bit larger? so the lines remain the same thickness, but have more clickable "padding" around them.
> 4. If I click the center of the textbox, it seems to jump back to the upper left corner.
> 5. The anchors (the red lines) seem to turn on when I drag an item around. They should really only be activated by the user to "lock" their relationship in place to the edges of the window.
> 6. This may be a bit tough, but if it's possible to update the numeric values in real-time, that would be awesome, but that could be a v2 feature.

Thanks for reviewing my code. I will update these feedbacks.

apps/studio/src/lib/editor/engine/move/index.ts
```diff
@@ -23,16 +23,28 @@ export class MoveManager {
...
-        this.originalIndex = await webview.executeJavaScript(
-            `window.api?.startDrag('${el.domId}')`,
-        );
 
-        if (this.originalIndex === null || this.originalIndex === -1) {
-            this.clear();
-            console.warn('Start drag failed, original index is null or -1');
-            return;
+        if (isAbsolute) {
+            this.originalIndex = await webview.executeJavaScript(
```
Kitenite, line 36: startTransaction here

Kitenite: https://github.com/onlook-dev/onlook/blob/main/apps/studio/src/lib/editor/engine/history/index.ts#L142

apps/studio/src/lib/editor/engine/move/index.ts
```diff
@@ -23,16 +23,28 @@ export class MoveManager {
             console.warn('Element not selected, cannot start drag');
             return;
         }
+
+        const computedStyle: Record<string, string> = await webview.executeJavaScript(
```
Kitenite, line 27: Use el: DomElement

apps/studio/src/lib/editor/engine/move/index.ts
```diff
@@ -77,27 +89,43 @@ export class MoveManager {
...
+            const newChild: DomElement | null = await webview.executeJavaScript(
+                `window.api?.endDragAbsoluteElement('${this.dragTarget.domId}')`,
+            );
+            if (!newChild) {
+                return;
+            }
+            this.editorEngine.style.updateMultiple({
+                left: newChild.styles?.computed?.left || '0px',
+                top: newChild.styles?.computed?.top || '0px',
+            });
```
Kitenite, line 104: commit transaction here

Kitenite: https://github.com/onlook-dev/onlook/blob/main/apps/studio/src/lib/editor/engine/history/index.ts#L146

apps/studio/src/routes/editor/EditPanel/StylesTab/compound/PositionInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+            return;
+        }
+        if (isCentered) {
+            return;
+        }
+        const { element, parent } = elements;
+        const centerX = (parent.rect.width - element.rect.width) / 2;
+        const centerY = (parent.rect.height - element.rect.height) / 2;
+
+        editorEngine.history.startTransaction();
```
Kitenite, line 145: No need for this

apps/studio/src/routes/editor/EditPanel/StylesTab/compound/PositionInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+    };
+
+    const checkIfCentered = useCallback(async () => {
+        const elements = await getElementAndParent();
+        if (!elements) {
+            return false;
+        }
+
+        const { element, parent } = elements;
+        const centerX = (parent.rect.width - element.rect.width) / 2;
```
Kitenite, line 160: image

See about doing calc the way framer is doing here

Kitenite: framer.com

apps/studio/src/lib/editor/engine/insert/index.ts
```diff
@@ -41,6 +41,7 @@ export class InsertManager {
                         width: '100px',
                         height: '100px',
                         backgroundColor: colors.blue[100],
+                        position: 'absolute',
```
Kitenite, line 44: Don't think this should be there

apps/studio/electron/preload/webview/elements/move/drag.ts
```diff
@@ -67,6 +68,9 @@ export function endDrag(domId: string): {
         return null;
     }
 
+    const newChild = getDomElement(el, true);
```
Kitenite, line 71: Why is this now true? It's false nefore `getDomElement(el, false),`

apps/studio/src/lib/editor/engine/move/index.ts
```diff
@@ -64,7 +85,7 @@ export class MoveManager {
     }
 
     async end(e: React.MouseEvent<HTMLDivElement>) {
-        if (this.originalIndex === undefined || !this.dragTarget) {
```
Kitenite, line 67: This check should still happen `this.originalIndex === undefined` when not absolute

apps/studio/src/lib/editor/engine/style/index.ts
```diff
@@ -36,9 +36,10 @@ export class StyleManager {
     }
 
     updateMultiple(styles: Record<string, string>) {
+        this.updateStyleNoAction(styles);
+        this.editorEngine.overlay.refreshOverlay();
```
Kitenite, line 40: Why do this? To refresh the animation?

apps/studio/src/lib/editor/engine/style/index.ts
```diff
@@ -89,6 +91,65 @@ export class StyleManager {
         };
     }
 
+    private calculateNewRects(
```
Kitenite, line 94: This is unused?

apps/studio/src/lib/editor/engine/style/index.ts
```diff
@@ -89,6 +91,65 @@ export class StyleManager {
...
+                rect.y = parentRect.y + (parseFloat(styles.top) || 0);
+            }
+
+            updates.set(domId, { rect, parentRect });
+        }
+
+        return updates;
+    }
+
+    private updateStyleAndRectNoAction(
```
Kitenite, line 123: Unused?

apps/studio/src/lib/editor/styles/group.ts
```diff
@@ -4,6 +4,10 @@ import { CompoundStyleKey, StyleType } from './models';
 import { ELEMENT_STYLE_UNITS } from './units';
 
 const STYLE_CONSTRAINTS = {
+    position: {
+        min: -9999,
```
Kitenite, line 8: Should there even be a constraint?

apps/studio/src/lib/editor/engine/insert/index.ts
spartan-vutrannguyen: image
It was my thought when i saw this. But you're right, let me remove it

PR #1439, opened 2025-02-19
Update .debug.script.mjs

arfaWong:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release

### DIFF
--- a/.vscode/.debug.script.mjs
+++ b/.vscode/.debug.script.mjs
@@ -4,7 +4,7 @@ import { fileURLToPath } from 'node:url'
 import { createRequire } from 'node:module'
 import { spawn } from 'node:child_process'
 
-const pkg = createRequire(import.meta.url)('../package.json')
+const pkg = createRequire(import.meta.url)('../apps/studio/package.json')
 const __dirname = path.dirname(fileURLToPath(import.meta.url))
 
 // write .debug.env


.vscode/.debug.script.mjs
```diff
@@ -4,7 +4,7 @@ import { fileURLToPath } from 'node:url'
 import { createRequire } from 'node:module'
 import { spawn } from 'node:child_process'
 
-const pkg = createRequire(import.meta.url)('../package.json')
+const pkg = createRequire(import.meta.url)('../apps/studio/package.json')
```
Kitenite, line 7: @arfaWong, what's the reasoning for this change? I'm not familiar with what this script does exactly as I'm not on vscode

arfaWong: https://github.com/onlook-dev/onlook/blob/main/.vscode/.debug.script.mjs#L11

arfaWong: There is no `pkg.debug.env` in the root package.json file but in apps/studio/package.json.

Will got TypeError: Cannot read properties of undefined (reading 'env') on
https://github.com/onlook-dev/onlook/blob/main/.vscode/.debug.script.mjs#L11

Kitenite: Gotcha thanks for fixing this

PR #1478, opened 2025-02-27
feat: watch file changes in directory

spartan-vutrannguyen:
## Description

- Watch file changes in directory
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release


apps/studio/electron/main/run/index.ts
```diff
@@ -126,41 +125,58 @@ class RunManager {
...
         if (this.watcher) {
             this.watcher.close();
             this.watcher = null;
         }
 
-        this.watcher = watch(filePaths, {
+        this.watcher = watch(projectDir, {
             persistent: true,
+            ignoreInitial: true,
+            ignored: [
```
Kitenite, line 137: Anything that's not jsx/tsx should be ignored as well.

apps/studio/electron/main/run/index.ts
```diff
@@ -50,8 +50,7 @@ class RunManager {
 
             this.setState(RunState.SETTING_UP, 'Setting up...');
             this.mapping.clear();
-            const filePaths = await this.addIdsToDirectoryAndCreateMapping(folderPath);
-            await this.listen(filePaths);
```
Kitenite, line 54: I would not change this. Please keep the listener outside of `addIdsToDirectoryAndCreateMapping` for separation of concern

apps/studio/electron/main/run/index.ts
```diff
@@ -126,41 +125,58 @@ class RunManager {
...
-    }
-
-    async addIdsToDirectoryAndCreateMapping(dirPath: string): Promise<string[]> {
+    async addIdsToDirectoryAndCreateMapping(dirPath: string): Promise<void> {
         const filePaths = await getValidFiles(dirPath);
         for (const filePath of filePaths) {
             await this.processFileForMapping(filePath);
         }
-        return filePaths;
+        await this.listen(dirPath);
```
Kitenite, line 179: Should keep the listen call on the outside like before

apps/studio/electron/main/run/index.ts
spartan-vutrannguyen: So instead of 'ignored', we should use 'allowed', right?

Kitenite: I see that it's filtered under 160 but this could save some compute if possible. If not that's pl

Kitenite: It should ignore the directories listed but should allow only jsx/tsx files. If possible. Otherwise change event will fire more than necessary

apps/studio/electron/main/run/index.ts
```diff
@@ -134,27 +134,34 @@ class RunManager {
 
         this.watcher = watch(filePaths, {
             persistent: true,
+            ignoreInitial: true,
+            ignored: (path: string) => {
+                if (ALLOWED_EXTENSIONS.some((ext) => path.endsWith(ext))) {
```
Kitenite, line 139: Still needs to check for IGNORED_DIRECTORIES as well

spartan-vutrannguyen: Why should we do this since we only allow files in ALLOWED_EXTENSIONS?

Kitenite: because you can have tsx/jsx files in node_modules or .next folders which we don't want to watch

Kitenite: The point is to only listen for code jsx/tsx files that actually matter such as the ones you would render on the page with.

apps/studio/electron/main/run/index.ts
```diff
@@ -126,41 +126,52 @@ class RunManager {
...
+                pollInterval: 100,
+            },
+            followSymlinks: false,
+            usePolling: false,
+            atomic: true,
         });
 
         this.watcher
             .on('change', (filePath) => {
-                this.processFileForMapping(filePath);
```
Kitenite, line 141: If you console log here, it still goes in an infinite loop. Are you sure you tested this?

spartan-vutrannguyen: Yesterday I tested it. Let me test again

Kitenite (changes requested):
Still one issue, change is being printed many times for any change instead of once. This is different from main where it's only triggered once. 
This seems to be an issue with addWatcher adding files that have already been listened to. Commenting out the add watcher makes it only print once so it seems to be a dir adding issue.

image

Kitenite (approved):
Not a fan of the arbitrary 500ms wait but this is ok

PR #1524, opened 2025-03-04
feat: read tailwind config and display

spartan-vutrannguyen:
## Description

- Read tailwind config from "tailwind.config.*" and "globals.css", display to the brand panel
- Allow update value and name of color.
- Allow Create/Update/Delete a group of colors

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Release

## Additional Notes

----

> [!IMPORTANT]
> Add functionality to read and manipulate Tailwind CSS configurations, including color operations, and introduce new IPC channels for these tasks.
> 
>   - **Behavior**:
>     - Read Tailwind config from `tailwind.config.*` and `globals.css`, display in brand panel.
>     - Support Create/Update/Delete operations for color groups and individual colors.
>   - **Functions**:
>     - `getConfigPath()`, `extractObject()`, `initializeTailwindColorContent()` in `helpers.ts` for config path and content handling.
>     - `updateTailwindColorConfig()`, `deleteTailwindColorGroup()` in `styles.ts` for color operations.
>   - **IPC Channels**:
>     - Add `SCAN_TAILWIND_CONFIG`, `UPDATE_TAILWIND_CONFIG`, `DELETE_TAILWIND_CONFIG` to `MainChannels` in `ipc.ts`.
>   - **Refactoring**:
>     - Move image handling functions from `index.ts` to `images.ts`.
>     - Rename and reorganize components in `ColorPanel` and `FontPanel` directories.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 84444df5fc8f81c0447ba4b47571b95a1b2c7215. It will automatically update as commits are pushed.</sup>


spartan-vutrannguyen:
https://github.com/user-attachments/assets/96e6d669-916f-4163-a223-14636e37e760

Two things I’d like to discuss:

- How should we handle color display for light and dark themes?
- When renaming a color, should we keep the old one for backward compatibility, or should we write a function to update the entire codebase?
@Kitenite @drfarrell

Kitenite:
@spartan-vutrannguyen 
> How should we handle color display for light and dark themes?

Could we add a new group or prefix? @tycreated or @drfarrell might have stronger opinions.  

> When renaming a color, should we keep the old one for backward compatibility, or should we write a function to update the entire codebase?

I think if we're able to update the codebase that may be better. Perhaps we can surface that as an option on Finish rename? Whether to update the existing usages or not. I like the idea of the variable changing usages on updates. 

> Currently, we can convert from hsl to hex. But how about hsla, should we try to convert to hex or keep it original format?

We should be able to use alpha hex. I believe the current color helpers already supports this.  But lmk if I'm not understanding correctly.

apps/studio/electron/main/assets/styles.ts
```diff
@@ -0,0 +1,780 @@
...
+    newColor: string,
+    theme?: 'dark' | 'light',
+): string {
+    // Constants for CSS selectors and patterns
+    const CSS_ROOT_SELECTOR = ':root';
+    const CSS_DARK_SELECTOR = '.dark';
+    const CSS_LAYER_BASE = '@layer base';
+
+    // More specific and readable regex patterns
+    const lightVarRegex = new RegExp(
```
spartan-vutrannguyen, line 328: Lots of Regex here. Should I use 'postcss' instead? @Kitenite

Kitenite: Yes, you should use css-tree or postcss instead. We use css-tree in other places in the app, specifically for applying stylesheets 
https://github.com/csstree/csstree
https://github.com/onlook-dev/onlook/blob/0fce59f4e5a674131e0f3ab0d79f96ad0537b614/apps/studio/electron/preload/webview/style/index.ts#L2

apps/studio/package.json
```diff
@@ -119,7 +119,7 @@
         "glob": "^11.0.1",
         "mobx": "^6.12.4",
         "mobx-react-lite": "^4.0.7",
-        "postcss": "^8.4.38",
+        "postcss": "^8.5.3",
```
Kitenite, line 122: I see you updated postcss but didn't use it?

Kitenite:
Looks good to me besides the regex stuff but we talked about that. Nice work!

PR #1526, opened 2025-03-05
fix: can not undo position change

spartan-vutrannguyen:
## Description

- After change position type, can not undo this.
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release


apps/studio/src/lib/editor/engine/move/index.ts
```diff
@@ -76,16 +76,31 @@ export class MoveManager {
...
+        y: number,
+    ) {
         const initialOffset = {
             x: dragOrigin.x - dragTarget.rect.x,
             y: dragOrigin.y - dragTarget.rect.y,
         };
 
-        const parentRect = dragTarget.parent?.rect;
-        if (!parentRect) {
+        const webview = this.editorEngine.webviews.getWebview(dragTarget.webviewId);
```
Kitenite, line 90: This also handles the issue. The parent isn't always the anchor element, it has to be relative or absolute positioned as well. Which is what causes this bug below. Can be fixed by getting the `offsetParent`

https://css-tricks.com/absolute-relative-fixed-positioining-how-do-they-differ/
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent

Clipboard-20250305-063741-278

PR #1565, opened 2025-03-08
fix: text color detection bug fixed

### LINKED ISSUE #1528: [bug] Style panel doesn't recognize text color
#### Describe the bug
When a text element is selected, the color setting displays "none" instead of the actual color of the text.
Image

0x23d11:
## Description

When selecting text elements in the editor (particularly headings), the properties panel was displaying text color as #000000 (black) even when the actual text had a different color (like red). This created a disconnect between what users saw in the editor and what was shown in the color properties panel.

My Approach:

Modified the `ColorInput` component to better handle text elements `(p, h1-h6, span, div)` by prioritizing the computed color values directly from the DOM, ensuring that both inherited and applied colors are displayed correctly.

Refined the `isColorEmpty` utility function to more accurately determine when a color is truly empty, preventing valid color values (such as black) from being incorrectly treated as empty.

## Related Issues

closes #1528 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release

https://drive.google.com/file/d/1e8t2h5EoVcqeLmFiZyMKZ_ft6KkSCEBH/view?usp=drive_link

### DIFF
--- a/apps/studio/.env.example
+++ b/apps/studio/.env.example
@@ -1,7 +1,7 @@
 # Only add these if you have valid keys. Otherwise, comment them out.
 VITE_SUPABASE_API_URL=
 VITE_SUPABASE_ANON_KEY=
-VITE_MIXPANEL_TOKEN=
+# VITE_MIXPANEL_TOKEN=
 
 # Add your keys here to use Anthropic directly
 VITE_ANTHROPIC_API_KEY=
--- a/apps/studio/src/lib/editor/engine/style/index.ts
+++ b/apps/studio/src/lib/editor/engine/style/index.ts
@@ -108,10 +108,51 @@ export class StyleManager {
         const newMap = new Map<string, SelectedStyle>();
         let newSelectedStyle = null;
         for (const selectedEl of selectedElements) {
-            const styles = {
-                ...selectedEl.styles?.computed,
-                ...selectedEl.styles?.defined,
-            };
+            // Debug log to check what styles are coming from the DOM
+            console.log('Selected element:', selectedEl.tagName);
+            console.log('Computed styles:', selectedEl.styles?.computed);
+            console.log('Defined styles:', selectedEl.styles?.defined);
+
+            // For text styles like color, prioritize the computed styles from the browser
+            // which contain the actual rendered values, especially for properties with inheritance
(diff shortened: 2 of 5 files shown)


packages/utility/src/color.ts
```diff
@@ -4,7 +4,16 @@ import parseCSSColor from 'parse-css-color';
...
+    // Check for common empty color values
+    if (!colorValue || colorValue === '' || colorValue === 'none' || colorValue === 'transparent') {
+        return true;
+    }
+
+    // Parse the color to check if it's truly transparent
     const color = Color.from(colorValue);
+
+    // Only consider it empty if it has zero alpha or exactly equals transparent
+    // Don't consider black (#000000) or other valid colors as empty
```
Kitenite, line 16: We actually don't consider black empty color. You might be seeing #00000000 which is transparent black. The issue seen is actually that there is no set color on the text color instead of it accidentally seeing black as empty. For example, if you set actual black #000000, it'll show up 
Screenshot 2025-03-08 at 5 17 24 PM

apps/studio/src/lib/editor/engine/style/index.ts
```diff
@@ -108,10 +108,51 @@ export class StyleManager {
         const newMap = new Map<string, SelectedStyle>();
         let newSelectedStyle = null;
         for (const selectedEl of selectedElements) {
-            const styles = {
-                ...selectedEl.styles?.computed,
-                ...selectedEl.styles?.defined,
-            };
+            // Debug log to check what styles are coming from the DOM
+            console.log('Selected element:', selectedEl.tagName);
```
Kitenite, line 112: We shouldn't ship logs in production

apps/studio/.env.example
```diff
@@ -1,7 +1,7 @@
 # Only add these if you have valid keys. Otherwise, comment them out.
 VITE_SUPABASE_API_URL=
 VITE_SUPABASE_ANON_KEY=
-VITE_MIXPANEL_TOKEN=
+# VITE_MIXPANEL_TOKEN=
```
Kitenite, line 4: I would not comment out the example here

PR #1666, opened 2025-03-24
feat: font panel config

spartan-vutrannguyen:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor


Kitenite (commented):
Is there a way to display the actual fonts? If we're already using the same next.js fonts package can we use it? We can skip this for custom fonts. 

Screenshot 2025-03-26 at 6 42 48 PM

apps/studio/src/routes/editor/LayersPanel/BrandTab/FontPanel/FontFamily.tsx
```diff
@@ -21,29 +21,34 @@ const FontVariant = ({ name, isActive = false }: FontVariantProps) => (
 export interface FontFamilyProps {
     name: string;
     variants: string[];
+    onRemoveFont?: () => void;
```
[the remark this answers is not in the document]
spartan-vutrannguyen: Because we reuse the FontFamily for both Site Fonts and System Fonts. System Fonts don't allow deleting so I made this optional

apps/studio/electron/main/assets/fonts.ts
```diff
@@ -0,0 +1,1012 @@
+import * as fs from 'fs';
```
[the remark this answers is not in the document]
spartan-vutrannguyen: Sure, I will split them into smaller chunks

apps/studio/src/routes/editor/LayersPanel/BrandTab/FontPanel/FontFamily.tsx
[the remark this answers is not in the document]
Kitenite: I see, that makes sense

drfarrell:
Some feedback on just the normal google font experience, not the custom fonts experience: 

**1. For a project I already had a Google font loaded into, it seems like it wasn't detected by the side panel. I wonder if it's because when I asked Onlook to add the font, it used next/font/google to bring it in. I think Next.js may do something special for fonts.** 

`'use client';

import { useState, useEffect } from 'react';
import { Figtree, PT_Serif } from 'next/font/google';

const figtree = Figtree({
    subsets: ['latin'],
    display: 'swap',
    variable: '--font-figtree',
});

const ptSerif = PT_Serif({
    weight: ['400', '700'],
    subsets: ['latin'],
    display: 'swap',
    variable: '--font-pt-serif',
});
`
When I first started the project it used Inter (I think that's our default) but FigTree and PT Serif are fonts that I asked Onlook's Chat to add. They don't show up in the "Added Fonts" section of the font panel. 

image

Here are screenshots of the files: 

Screenshot 2025-03-30 at 10 18 29 PM

Screenshot 2025-03-30 at 10 18 55 PM

Screenshot 2025-03-30 at 10 19 08 PM

I know this can be a bit of an edge-case situation, but ideally the installed fonts also show up in the "Added Fonts" and can be controlled from the left panel. If this explodes the scope of this PR too much, I'm happy to push it to another PR. 


**2. I loaded in a font from Google Fonts (Andada Pro), and it made it in...**

Screenshot 2025-03-30 at 10 24 43 PM

When I set Inter as the default, it looks like this: 
Screenshot 2025-03-30 at 10 25 06 PM

Screenshot 2025-03-30 at 10 25 27 PM

But when I set Andada Pro as the default, it looks like this: 
image

The code seems to still include Inter. 
Screenshot 2025-03-30 at 10 26 05 PM

It seems like a few things are happening....
- The ${inter.variable}  should be  ${andadaPro.variable}
- The import at the top of the page shouldn't just be import { inter } from '@/app/fonts'; but import { andadaPro, inter } from '@/app/fonts';

I used Cursor Tab to make these suggestions. When I used Cursor tab, it correctly imported the font, and now it looks like this (correct):
image

Ok feedback on the custom font experience: 

**3. Seems to work well just loading the font in. We'll likely need to design how selecting an italic for a font would work, but otherwise good**

**4. Small style improvement**
Add cursor-pointer class to the dropdown and give it the same hover-state as the trash can on the right so it feels more "clickable"
image

**5. Font panel breaks for long text**

image

Let's make the width fixed for the labels, so in this case it would be...

customCassannetPlusW.. (Default) [...] 

image

Then when the user hovers over the name, I'd like a tooltip off-to-the-side with the full-name, just like the layers. You can borrow the max-width styling and such from there.

image

apps/studio/electron/main/assets/fonts/font.ts
```diff
@@ -0,0 +1,414 @@
...
+            if (newContent.endsWith('\n')) {
+                newContent += '\n';
+            } else {
+                newContent += '\n\n';
+            }
+        }
+
+        newContent += fontConfig;
+
+        fs.writeFileSync(fontPath, newContent);
```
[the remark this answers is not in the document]
Kitenite: @spartan-vutrannguyen , writeFileSync is discouraged. Please change to the promisified writeFile

apps/studio/electron/main/assets/fonts/font.ts
```diff
@@ -0,0 +1,371 @@
...
+        await addFont(projectRoot, font);
+    }
+}
+
+/**
+ * Adds a local font to the project by:
+ * 1. Saving the font files to the fonts folder
+ * 2. Adding the font configuration to fonts.ts using next/font/local
+ */
+export async function addLocalFont(
```
Kitenite, line 269: Same here, should use Babel/AST which you are using for removeFont

apps/studio/electron/main/assets/fonts/font.ts
```diff
@@ -0,0 +1,346 @@
...
+                const style = fontFile.style.toLowerCase();
+                const fileName = getFontFileName(baseFontName, weight, style);
+                const filePath = pathModule.join(
+                    fontsDir,
+                    `${fileName}.${fontFile.file.name.split('.').pop()}`,
+                );
+
+                // Save the file
+                const buffer = Buffer.from(fontFile.file.buffer);
+                fs.writeFileSync(filePath, buffer);
```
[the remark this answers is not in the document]
Kitenite: Same here, should await writeFile instead

apps/studio/electron/main/assets/fonts/tailwind.ts
```diff
@@ -0,0 +1,154 @@
...
+                            return true; // Signal that we updated the config
+                        }
+                    }
+                }
+                return false;
+            },
+        });
+
+        if (isUpdated) {
+            fs.writeFileSync(configPath, output);
```
Kitenite, line 144: Pls use awaited writeFile here as well.

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        this._fontFamilies = fontFamilies.map((font) => ({
+            ...font,
+            weight: font.weights.map((weight) => weight.toString()),
+            styles: font.styles.map((style) => style.toString()),
+            variable: `--font-${font.id}`,
+        }));
+    }
+
+    private async watchFontFile(projectRoot: string) {
+        console.log('Setting up font file watcher for path:', projectRoot);
```
Kitenite, line 47: Leftover console logs?

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (!projectRoot) {
+            console.log('No project root provided, skipping font file watcher setup');
```
Kitenite, line 49: Should be error

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+            return;
+        }
+
+        try {
+            console.log('Invoking WATCH_FONT_FILE with project root:', projectRoot);
+            await invokeMainChannel(MainChannels.WATCH_FONT_FILE, {
+                projectRoot,
+            });
+            console.log('Font file watcher setup complete');
```
Kitenite, line 58: Logs

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    private async watchFontFile(projectRoot: string) {
+        console.log('Setting up font file watcher for path:', projectRoot);
+        if (!projectRoot) {
+            console.log('No project root provided, skipping font file watcher setup');
+            return;
+        }
+
+        try {
+            console.log('Invoking WATCH_FONT_FILE with project root:', projectRoot);
```
Kitenite, line 54: logs

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+            console.log('Font file watcher setup complete');
+        } catch (error) {
+            console.error('Error setting up font file watcher:', error);
+        }
+    }
+
+    async scanFonts() {
+        const projectRoot = this.projectsManager.project?.folderPath;
+        if (!projectRoot) {
+            return;
```
Kitenite, line 67: Should console.error here

apps/studio/src/lib/editor/engine/font/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        })) as Font[];
+
+        this._fonts = fonts;
+        return fonts;
+    }
+
+    async addFont(font: Font) {
+        const projectRoot = this.projectsManager.project?.folderPath;
+        if (!projectRoot) {
+            return;
```
Kitenite, line 81: Same with these

Kitenite (commented):
Mostly stylistic comments and some cleanups. I think it's pretty close.

spartan-vutrannguyen:
Great suggestions, @drfarrell ! I appreciate the insights and  I’ve incorporated them into the latest commit. Thanks for taking the time to review!

> Add cursor-pointer class to the dropdown and give it the same hover-state as the trash can on the right so it feels more "clickable"

> Font panel breaks for long text

> Suggest font weights from font-names

https://github.com/user-attachments/assets/590246ea-a359-4764-ac3c-a7de9701ee8b

Kitenite (approved):
Awesome nice work on this!

PR #1669, opened 2025-03-24
Enable brand panel

Kitenite:
## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Enable brand panel by activating `BrandPopoverPicker` in `ColorInput` and making brand tab visible in `LayersPanel`.
> 
>   - **Behavior**:
>     - Enable `BrandPopoverPicker` in `ColorInput` by uncommenting its usage in `index.tsx`.
>     - Make brand tab visible in `LayersPanel` by removing `hidden` class from the button in `index.tsx`.
>   - **Misc**:
>     - Comment out `PopoverPicker` in `ColorInput` in `index.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 6a1046f12431f09efdbfcf9fba3575405f222ef8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/studio/src/routes/editor/EditPanel/StylesTab/single/ColorInput/index.tsx
+++ b/apps/studio/src/routes/editor/EditPanel/StylesTab/single/ColorInput/index.tsx
@@ -7,8 +7,7 @@ import { Icons } from '@onlook/ui/icons';
 import { Color, isColorEmpty } from '@onlook/utility';
 import { observer } from 'mobx-react-lite';
 import { memo, useCallback, useEffect, useMemo, useState } from 'react';
-// import { BrandPopoverPicker } from './ColorBrandPicker';
-import { PopoverPicker } from './Popover';
+import { BrandPopoverPicker } from './ColorBrandPicker';
 
 const stripUrlWrapper = (url: string) => {
     return url.replace(/^url\((['"]?)(.*)\1\)/, '$2');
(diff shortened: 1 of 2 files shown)


thughey:
1. Renaming color group needs to update on save:
https://github.com/user-attachments/assets/c525cb95-5839-4902-99eb-f14d46b94fb4

2. Can we save + close the modal after pressing enter/return?

3. If a color name is set with a space can we display the space in the context menu & in the name input?
CleanShot 2025-03-24 at 17 13 32@2x
CleanShot 2025-03-24 at 17 15 08@2x

4. Editing the color in light or darkmode ends up updating both to the same color:
https://github.com/user-attachments/assets/9101806e-671e-4d17-b59a-4c459526cea6

drfarrell:
1. If I hit "enter" after choosing a new color, it should close the panel and set the color, essentially overriding the "enter to text" functionality
image

Enter to text functionality: 
image

2. CMD+Z should also apply to a color change. So rather than undoing what happened in the editor, it should undo the color change I made in the brand panel.

3. If I change the color setting of a pre-existing color, it should update the color live (OR at the very least refresh the page when I'm done so I can see the color change). (Perhaps this is a me thing, I can try a new project though)

4. Prevent spaces or breaking characters from being used as color variable names.
image

5. When you click "Rename" for a group, auto-focus the input field so the user doesn't have to click it.
image

PR #1936, opened 2025-05-27
refactor: user specific canvas state

satya-nutella:
# Implement User-Specific Canvas State Management

## Description

Refactored canvas data storage to be user-specific, allowing multiple users to have independent canvas view states (scale, pan position) for the same project. This enables proper multi-user collaboration where each user maintains their own viewport settings.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Release
- [x] Refactor

### Database Schema
- **Created `user_canvases` table** with composite primary key (`user_id`, `canvas_id`) to store user-specific canvas viewport data
- **Added foreign key constraints** linking to `users` and `canvas` tables with cascade delete/update
- **Enabled row-level security** on the new table
- **Migrated existing canvas data** from `canvas` table to `user_canvases` via JOIN with `user_projects`
- **Removed viewport columns** (`scale`, `x`, `y`) from `canvas` table as they're now user-specific

### API Layer
- **Added `userCanvasRouter`** with get/update endpoints for user canvas operations
- **Updated project router** to fetch and return both canvas and user canvas data
- **Modified canvas router** to return raw canvas data without viewport information
- **Implemented user canvas utilities** for creating default entries

### Frontend Integration
- **Updated canvas manager** to use user canvas API endpoints instead of general canvas endpoints
- **Modified project data flow** to apply user canvas data to editor engine
- **Added user canvas creation** during project setup to ensure default entries exist

### Utility Functions
- **Created `createDefaultUserCanvas`** helper function for generating default user canvas entries
- **Updated canvas creation** to exclude viewport data (now handled by user canvas)

## Testing

- Verify existing projects maintain their canvas viewport settings after migration
- Test that new users get default canvas settings when accessing projects
- Confirm multiple users can have independent viewport states for the same project
- Validate cascade deletes work properly when users or canvases are removed

## Additional Notes

This change maintains backward compatibility through the migration script that preserves existing canvas viewport data. The refactor enables true multi-user collaboration by isolating each user's canvas view state while maintaining shared project and frame data.

----

> [!IMPORTANT]
> Refactor to implement user-specific canvas state management, enabling independent viewport settings per user.
> 
>   - **Database Schema**:
>     - Create `user_canvases` table with composite primary key (`user_id`, `canvas_id`) for user-specific viewport data.
>     - Add foreign key constraints to `users` and `canvas` tables with cascade delete/update.
>     - Enable row-level security on `user_canvases`.
>     - Migrate viewport data from `canvas` to `user_canvases`.
>     - Remove `scale`, `x`, `y` columns from `canvas` table.
>   - **API Layer**:
>     - Add `userCanvasRouter` with get/update endpoints in `user-canvas.ts`.
>     - Update `project.ts` to handle user canvas data.
>     - Modify `canvas.ts` to return raw canvas data without viewport info.
>   - **Frontend Integration**:
>     - Update `main.tsx` to use user canvas data.
>     - Modify `CanvasManager` in `canvas/index.ts` to update user canvas.
>   - **Utility Functions**:
>     - Create `createDefaultUserCanvas` in `user-canvas.ts` for default user canvas entries.
>     - Update `createDefaultCanvas` to exclude viewport data.
>   - **Testing**:
>     - Ensure projects maintain viewport settings post-migration.
>     - Verify new users receive default canvas settings.
>     - Confirm independent viewport states for multiple users.
>     - Validate cascade deletes for users/canvases.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 66c79bc81ff210f251c4996a6cc0d6a237a4861e. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/web/client/src/components/store/editor/canvas/index.ts
+++ b/apps/web/client/src/components/store/editor/canvas/index.ts
@@ -64,9 +64,7 @@ export class CanvasManager {
     }
 
     async updateCanvas(canvas: Canvas) {
-        const success = await api.canvas.update.mutate(
-            fromCanvas(this.projects.project?.id ?? '', canvas),
-        );
+        const success = await api.userCanvas.update.mutate(fromCanvas(canvas));
         if (!success) {
             console.error('Failed to update canvas');
         }
@@ -95,13 +93,11 @@ export class CanvasManager {
 
     private undebouncedSaveSettings() {
         if (this.projects.project) {
-            this.updateCanvas(
-                {
-                    id: this.id,
-                    position: this.position,
-                    scale: this.scale,
-                },
-            );
+            this.updateCanvas({
+                id: this.id,
+                position: this.position,
+                scale: this.scale,
+            });
         }
     }
 }
--- a/apps/web/client/src/server/api/routers/user-canvas.ts
+++ b/apps/web/client/src/server/api/routers/user-canvas.ts
@@ -17,22 +17,20 @@ export const userCanvasRouter = createTRPCRouter({
             }),
         )
         .query(async ({ ctx, input }) => {
-            const result = await ctx.db
-                .select()
-                .from(canvases)
-                .innerJoin(userCanvases, eq(canvases.id, userCanvases.canvasId))
-                .where(
-                    and(
-                        eq(canvases.projectId, input.projectId),
-                        eq(userCanvases.userId, ctx.user.id),
-                    ),
-                )
-                .limit(1);
+            const userCanvas = await ctx.db.query.userCanvases.findFirst({
+                where: and(
+                    eq(canvases.projectId, input.projectId),
+                    eq(userCanvases.userId, ctx.user.id),
+                ),
+                with: {
+                    canvas: true,
+                },
+            });
 
-            if (!result[0]) {
+            if (!userCanvas) {
                 return null;
             }
-            return toCanvas(result[0].user_canvases);
+            return toCanvas(userCanvas);
         }),
     update: protectedProcedure.input(userCanvasUpdateSchema).mutation(async ({ ctx, input }) => {
         try {
--- a/apps/web/client/src/server/api/routers/user-canvas.ts
+++ b/apps/web/client/src/server/api/routers/user-canvas.ts
@@ -1,9 +1,8 @@
 import {
     canvases,
-    canvasUpdateSchema,
     toCanvas,
     userCanvases,
-    userCanvasUpdateSchema,
+    userCanvasUpdateSchema
 } from '@onlook/db';
 import { and, eq } from 'drizzle-orm';
 import { z } from 'zod';
@@ -28,7 +27,7 @@ export const userCanvasRouter = createTRPCRouter({
             });
 
             if (!userCanvas) {
-                return null;
+                throw new Error('User canvas not found');
             }
             return toCanvas(userCanvas);
         }),
--- a/apps/web/client/src/server/api/routers/user-canvas.ts
+++ b/apps/web/client/src/server/api/routers/user-canvas.ts
@@ -1,0 +1,57 @@
+import {
+    canvases,
+    canvasUpdateSchema,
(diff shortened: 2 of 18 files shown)


packages/db/src/dto/canvas.ts
```diff
@@ -1,23 +1,22 @@
...
-            x: Number(dbCanvas.x),
-            y: Number(dbCanvas.y),
+            x: Number(dbUserCanvas.x),
+            y: Number(dbUserCanvas.y),
         },
     };
 };
 
-export const fromCanvas = (projectId: string, canvas: Canvas): DbCanvas => {
+export const fromCanvas = (canvas: Canvas): Omit<DbUserCanvas, 'userId'> => {
```
Kitenite, line 15: Not sure about returning an omit type here. Depending on the use case perhaps we remove this function in general for something more specific/useful

apps/web/client/src/server/api/routers/user-canvas.ts
```diff
@@ -0,0 +1,57 @@
...
+import { createTRPCRouter, protectedProcedure } from '../trpc';
+
+export const userCanvasRouter = createTRPCRouter({
+    get: protectedProcedure
+        .input(
+            z.object({
+                projectId: z.string(),
+            }),
+        )
+        .query(async ({ ctx, input }) => {
```
Kitenite, line 19: Drizzle has a better syntax for resolving the relationship similar to here:
https://github.com/onlook-dev/onlook/blob/53b7f84552e9169d88f521c0ee17f4e80847be86/apps/web/client/src/server/api/routers/user.ts#L46
```
 const user = await ctx.db.query.users.findFirst({
            where: eq(users.id, input),
            with: {
                userProjects: {
                    with: {
                        project: true,
                    },
                },
            },
        });
        return user;
    }),
```
```
 const result = await ctx.db.query.userCanvases.findFirst({
                where: and(
                    eq(canvases.projectId, input.projectId),
                    eq(userCanvases.userId, ctx.user.id),
                ),
                with: {
                    canvas: true,
                },
            });

            if (!result) {
                console.log('No canvas found for user', ctx.user.id, 'and project', input.projectId);
                return null;
            }

            return toResolvedObject(result);
```

the returned type would then be which is nicer to work with

```
const result: {
    canvasId: string;
    x: string;
    y: string;
    userId: string;
    scale: string;
    canvas: {
        id: string;
        projectId: string;
    };
}
```

packages/db/src/seed/db.ts
```diff
@@ -1,4 +1,4 @@
-import { canvases, conversations, frames, messages, projects, userProjects, users, type Conversation, type Message, type Project, type User } from '@onlook/db';
+import { canvases, conversations, frames, messages, projects, userProjects, userCanvases, users, type Conversation, type Message, type Project, type User } from '@onlook/db';
```
Kitenite, line 1: we should probably update the seed for easier testing. just adding the userCanvas would help

apps/web/client/src/components/store/editor/canvas/index.ts
```diff
@@ -64,9 +64,7 @@ export class CanvasManager {
     }
 
     async updateCanvas(canvas: Canvas) {
-        const success = await api.canvas.update.mutate(
-            fromCanvas(this.projects.project?.id ?? '', canvas),
-        );
+        const success = await api.userCanvas.update.mutate(fromCanvas(canvas));
```
Kitenite, line 67: I think we should add the user ID into this mutation then so that we ensure we're updating the right userCanvas. 

Seems like there are 2 patterns. 
1. Use the ID for the user_canvas row. 
2. Use the canvasId and userId. 

Looks like this is using option 1 but we can talk about it.

PR #1973, opened 2025-05-30
feat: realtime rls

satya-nutella:
## Description

- adds comments back to the rls migrations
- adds a CI script to run migrations on supabase db on workflow dispatch
- adds triggers for rls on conversations table and messages table

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Release
- [x] Refactor


apps/backend/supabase/migrations/0007_realtime_rls.sql
```diff
@@ -0,0 +1,53 @@
+create or replace function public.project_changes()
```
Kitenite, line 1: Can we keep consistent capitalization format across the migrations?

PR #2025, opened 2025-06-04
feat: publish to domain

spartan-vutrannguyen:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Adds domain publishing feature with UI components, domain management logic, and utility functions for file operations and domain verification.
> 
>   - **Behavior**:
>     - Adds `PublishButton` and `PublishDropdown` components in `publish/index.tsx` and `publish/dropdown/index.tsx` for domain publishing UI.
>     - Implements `DomainSection` and `AdvancedSettingsSection` components for domain management in `publish/dropdown/domain.tsx` and `publish/dropdown/advanced-settings.tsx`.
>     - Supports publishing to base and custom domains with status updates (loading, published, error) in `domain.tsx`.
>   - **Domain Management**:
>     - Introduces `DomainsManager` in `domain.ts` for handling domain states and publishing logic.
>     - Adds `HostingManager` in `hosting.ts` for deployment logic, including file serialization and badge management.
>     - Implements domain verification and ownership checks in `domain.ts` and `utility/src/domain.ts`.
>   - **Utilities**:
>     - Adds `FileOperations` interface in `utility/src/file.ts` for abstracted file operations.
>     - Introduces `LogTimer` class in `utility/src/time.ts` for performance logging.
>     - Updates `constants` and `parser` packages for configuration management and file handling.
>   - **Misc**:
>     - Updates `package.json` dependencies to include `@onlook/utility` and `freestyle-sandboxes`.
>     - Changes default build commands to use `bun` in `constants/src/editor.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 82cff0a8629db5bd3a68ab2772bfa0fe3dacd9d3. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/src/app/layout.tsx
```diff
@@ -53,7 +53,7 @@ export default async function RootLayout({ children }: { children: React.ReactNo
         <html lang={locale} className={inter.variable} suppressHydrationWarning>
             <body>
                 <FeatureFlagsProvider>
-                    <PostHogProvider>
+                    {/* <PostHogProvider> */}
```
Kitenite: Why comment this out?

apps/web/client/src/components/store/editor/sandbox/session.ts
```diff
@@ -85,10 +89,29 @@ export class SessionManager {
                 throw new Error('No session found');
             }
 
-            const terminalSession = this.terminalSessions.get(this.activeTerminalSessionId) as TerminalSession | undefined;
+            // Try to get the active terminal session first
+            let terminalSession = this.terminalSessions.get(this.activeTerminalSessionId);
```
Kitenite: I don't think we should use require a terminal session. Instead we should just run the command. If there is a terminal session then we can write the output there.

apps/web/client/src/components/store/editor/sandbox/terminal.ts
```diff
@@ -46,12 +48,18 @@ export class CLISessionImpl implements CLISession {
...
+            this.initializationPromise = this.initTerminal();
         } else if (type === CLISessionType.TASK) {
-            this.initTask();
+            this.initializationPromise = this.initTask();
+        } else {
+            this.initializationPromise = Promise.resolve();
         }
     }
 
+    async waitForInitialization(): Promise<void> {
```
Kitenite: I'm not sure I like this pattern. This can hypothetically hang forever if we call this without an explicit timeout. Instead we should just handle if session is not initialized.

apps/web/client/src/components/store/project/manager.ts
```diff
@@ -1,32 +1,62 @@
...
+import type { EditorEngine } from '../editor/engine';
 
 export class VersionsManager {
     constructor(private projectManager: ProjectManager) { }
 }
 
 export class ProjectManager {
     private _project: Project | null = null;
-    readonly domains: DomainsManager | null = null;
+    private _domains: DomainsManager | null = null;
```
Kitenite: Could you consider if it would be better to keep domain manager outside of project manager? Perhaps as its own hook?  I don't like how stateful the whole project manager is.

apps/web/client/src/components/store/project/manager.ts
```diff
@@ -1,32 +1,62 @@
...
         this.versions = new VersionsManager(this);
         makeAutoObservable(this);
     }
 
+    setEditorEngine(editorEngine: EditorEngine) {
+        this._editorEngine = editorEngine;
+        this.updateDomainsManager();
+    }
+
+    private updateDomainsManager() {
```
Kitenite: This is the part I don't particularly like. It feels wrong to keep reinitializing with a project changing. Perhaps it's better to initialize once and have a reaction within domainsManager for the project changing.

packages/foundation/package.json
```diff
@@ -0,0 +1,50 @@
+{
+    "name": "@onlook/foundation",
```
Kitenite: I would not bring this package back tbh, can we move parser related stuff to parser?

packages/foundation/src/frameworks/next.ts
```diff
@@ -0,0 +1,345 @@
...
+ *   fileExists: (path: string) => editorEngine.sandbox.fileExists(path)
+ * };
+ *
+ * // Then use the functions:
+ * await isNextJsProject(fileOps);
+ * await modifyNextConfig('.js', fileOps);
+ * await addNextBuildConfig(fileOps);
+ * await removeNextCache(fileOps);
+ */
+import generate from '@babel/generator';
```
Kitenite: All this should be in parser. We can no longer support these babel dependencies

apps/web/client/src/app/layout.tsx
spartan-vutrannguyen: It does not work correctly on my laptop so I comment this out while developing

apps/web/client/.env.example
```diff
@@ -19,11 +19,16 @@ MORPH_API_KEY="<Your api key from https://morphllm.com/dashboard>"
 # Option 2: Relace
 RELACE_API_KEY="<Your api key from https://app.relace.ai/settings/api-keys>"
 
+# Freestyle - Deploy your website
+FREESTYLE_API_KEY="<Your api key from https://admin.freestyle.sh/"
```
Kitenite: Should update the env script and docs if you have time but can do later. 
should be optional keys.

apps/web/client/src/app/project/[id]/_components/top-bar/index.tsx
```diff
@@ -15,9 +15,12 @@ import { ModeToggle } from './mode-toggle';
 import { ProjectBreadcrumb } from './project-breadcrumb';
 import { Members } from '../members';
 import { useFeatureFlags } from '@/hooks/use-feature-flags';
+import { useProjectManager } from '@/components/store/project';
+import { Publish } from './publish';
 
 export const TopBar = observer(({ projectId }: { projectId: string }) => {
     const editorEngine = useEditorEngine();
+    const projectManager = useProjectManager();
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/top-bar/index.tsx]: this doesn't seem necessary?

apps/web/client/src/app/project/[id]/_components/top-bar/publish.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useEffect } from 'react';
+import { PublishDropdown } from './publish/dropdown';
+import { PublishButton } from './publish/trigger-button';
+
+export const Publish = observer(() => {
+    const userManager = useUserManager();
+    const editorEngine = useEditorEngine();
+
+    useEffect(() => {
+        userManager.subscription.getPlanFromServer();
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/top-bar/publish.tsx]: Maybe we can do a reaction in userManager.subscription instead of hooks here?

packages/constants/src/files.ts
```diff
@@ -14,3 +14,16 @@ export const IGNORED_DIRECTORIES = [
...
 export const JS_FILE_EXTENSIONS = ['js', 'ts'];
+
+export const SUPPORTED_LOCK_FILES = [
+    'bun.lock',
+    'package-lock.json',
+    'yarn.lock',
+    'pnpm-lock.yaml',
+];
+
+export enum FILE_EXTENSION {
```
Kitenite: perhaps we should merge this with JS_FILE_EXTENSIONS unless there's some weird usage reason we don't

packages/constants/src/files.ts
```diff
@@ -13,17 +13,11 @@ export const IGNORED_DIRECTORIES = [
 
 export const JSX_FILE_EXTENSIONS = ['jsx', 'tsx'];
 
-export const JS_FILE_EXTENSIONS = ['js', 'ts'];
+export const JS_FILE_EXTENSIONS = ['.js', '.ts', '.mjs'];
```
Kitenite: I would make sure that adding a '.' doesn't mess up every single usage of this that assumes the opposite

Breaks this for example: https://github.com/onlook-dev/onlook/blob/e2e777b2b440ba0ddd20b9551511f056bd58bb8d/apps/web/client/src/components/store/editor/sandbox/helpers.ts#L18

PR #2086, opened 2025-06-06
fix: aspect ratio bug on resize with locked ratio

### LINKED ISSUE #931: [bug] Window aspect ratio is not consistent
#### Describe the bug
If I have the aspect ratio locked in at 1, then resizing using any of the handles works perfectly. But if I have the aspect ratio locked at anything other than 1 then whenever I use the right or bottom handle the other dimension is not adjusted correctly every time.

Steps to reproduce: 

1. Lock the aspect ratio with width>height or vice versa so the aspect ratio is not 1
2. Try resizing the frame using the resize handles, and if you use the right handle, you'll notice that when you increase the width and decrease it back to the minimum width, the height doesn't come back to the exact height calculated according to the aspect ratio, sometimes it does, sometimes it doesn't. Same is the case with the bottom handle as well.

kishansinghifs1:
### What
Fixes #931 — properly adjusts the other dimension when resizing with locked aspect ratio ≠ 1.

### Why
Previously, resizing with aspect ratio ≠ 1 (e.g., 4:3) only worked from the corner, not from right or bottom handle.

### How
- Modified delta handling logic in ResizeHandles.tsx
- Synced width/height update according to the locked aspect ratio when resizing from single-axis handles

Tested with aspect ratios: 2:1, 4:3, 16:9, etc.

## Screen Recording Description for Bug #931

https://github.com/user-attachments/assets/10ac0f6d-5829-491c-a990-ac5d30a20d64

### DIFF
--- a/apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
+++ b/apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
@@ -3,7 +3,7 @@ import type { FrameImpl } from '@/components/store/editor/frames/frame';
 import { DefaultSettings } from '@onlook/constants';
 import { cn } from '@onlook/ui/utils';
 import { observer } from 'mobx-react-lite';
-import { type MouseEvent } from 'react';
+import { MouseEvent } from 'react';
 
 enum HandleType {
     Right = 'right',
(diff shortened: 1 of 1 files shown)


Kitenite (approved):
Awesome work! Thank @kishansinghifs1

PR #2101, opened 2025-06-07
fix: prevent infinite restart loop by lazy-loading FreestyleSandboxes…

### LINKED ISSUE #2099: [bug] bun dev command is stuck in an infinite restart loop!
#### Describe the bug
```
@onlook/web-client dev:   5 |
│ @onlook/web-client dev: > 6 | const sdk = new FreestyleSandboxes({
│ @onlook/web-client dev:     |            ^
│ @onlook/web-client dev:   7 |     apiKey: env.FREESTYLE_API_KEY
│ @onlook/web-client dev:   8 | });
│ @onlook/web-client dev:   9 | {
```
I noticed that the FREESTYLE_API_KEY environment variable is marked as optional in the schema `(z.string().optional())`, but it's still being used to instantiate FreestyleSandboxes in the domain router.

The issue is that if the env variable is missing or undefined, it ends up trying to create the SDK with undefined as the API key. That probably causes the constructor to throw, which then crashes the dev server and since it's in the startup path, it just keeps restarting in an infinite loop.

We’ll either need to make the env var required or add a proper check before initializing the SDK.

Rish-it:
fix: prevent infinite restart loop by lazy-loading FreestyleSandboxes SDK

## Description

Fixes infinite dev server restart loop caused by FreestyleSandboxes SDK initialization failure when `FREESTYLE_API_KEY` environment variable is not set.

**Problem**: The FreestyleSandboxes SDK was being instantiated at module import time, causing the dev server to crash and restart infinitely if the `FREESTYLE_API_KEY` environment variable was missing or undefined.

**Solution**: 
- Moved SDK instantiation from module-level to a lazy-loaded function (`createFreestyleSdk`)
- SDK is now only created when the domain publishing mutation is actually called
- Added proper error handling with clear messaging for missing API key
- Allows the application to start successfully even without the `FREESTYLE_API_KEY` environment variable

**Impact**: Developers can now run the dev server without configuring the Freestyle API key, and will only get an error if they actually try to use the domain publishing feature.

## Related Issues

Closes #2099

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Testing

- Verified dev server starts successfully without `FREESTYLE_API_KEY` set
- Confirmed domain publishing throws appropriate error when API key is missing
- Tested that domain publishing works correctly when API key is provided
- No breaking changes to existing functionality

Before: 
Screenshot 2025-06-07 at 6 15 39 PM

After:

https://github.com/user-attachments/assets/b5528a22-8ca6-4e22-8120-ed1cc28df66c

## Additional Notes
This fix resolves the exact issue described in #2099 where the `FREESTYLE_API_KEY` environment variable being optional in the schema but required at SDK instantiation time caused infinite restarts.

### DIFF
--- a/apps/web/client/src/server/api/routers/domain.ts
+++ b/apps/web/client/src/server/api/routers/domain.ts
@@ -2,10 +2,14 @@ import { env } from '@/env';
 import { FreestyleSandboxes, type FreestyleDeployWebSuccessResponseV2 } from 'freestyle-sandboxes';
 import { z } from 'zod';
 import { createTRPCRouter, protectedProcedure } from '../trpc';
+import { TRPCError } from '@trpc/server';
 
 const createFreestyleSdk = () => {
     if (!env.FREESTYLE_API_KEY) {
-        throw new Error('FREESTYLE_API_KEY environment variable is not set');
+        throw new TRPCError({
+            code: 'PRECONDITION_FAILED',
+            message: 'FREESTYLE_API_KEY environment variable is not set. Please configure it to use domain publishing features.',
+        });
     }
     return new FreestyleSandboxes({
         apiKey: env.FREESTYLE_API_KEY
--- a/apps/web/preload/dist/index.js
+++ b/apps/web/preload/dist/index.js
@@ -13248,10 +13248,122 @@ async function renderElement(context, element, rect, styles) {
     }
   }
 }
-// ../../../node_modules/zod/lib/index.mjs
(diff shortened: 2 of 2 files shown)


Kitenite:
Closing as duplicate to a similar fix. Thanks for this! Sorry about the issue.
https://github.com/onlook-dev/onlook/pull/2097

PR #2121, opened 2025-06-09
Fix/font search bug

### LINKED ISSUE #2058: [bug] Font search does not work as expected
**Summary:**
The font search feature returns incorrect results. When searching for a full font name (e.g., "Adlam Display"), the search returns any font that contains part of the term (e.g., fonts with just “Display”), instead of matching the full query string.

**Steps to reproduce:**
1. Open the font panel.
2. In the search input, type a full font name such as "Adlam Display".

**Actual result:**
Fonts that only contain "Display" appear in the results, including ones that do not contain "Adlam".
https://github.com/user-attachments/assets/5809fe60-7fd2-4189-893c-eeaad9ceabe2

**Expected result:**
Only fonts that match the full search string, such as "Adlam Display", should be shown in the results.

kishansinghifs1:
## Description

Fixes the font search logic to properly match full query strings. Previously, searching for a font like "Adlam Display" would return results that only partially matched the term (e.g., fonts that contained just "Display"), instead of requiring a more complete match.

The updated logic uses FlexSearch with suggest and enrich options, maps and filters results, and applies a stricter condition for multi-word queries to ensure full, case-insensitive matching.

## Related Issues

Fixes #2058 Font search doesn't work as expected .

## Type of Change

 Bug fix

## Testing

Verified the font search returns exact match results for multi-word queries like "Adlam Display".

Ensured partial matches still work for single-word queries like "Roboto".

Confirmed that no fonts are returned when no match is found.

Manually validated that _searchResults is updated only with valid results.

Checked that loadFontBatch() is only triggered when valid fonts are found.

No console errors were encountered during searches.

----

> [!IMPORTANT]
> Fixes font search logic for full query matches and updates UI with animations and style changes.
> 
>   - **Font Search Logic**:
>     - Fixes font search logic in `FontManager` to ensure full query string matches for multi-word queries using FlexSearch.
>     - Filters results for exact case-insensitive matches when query has multiple words.
>   - **UI Enhancements**:
>     - Updates `not-found.tsx` to include animations using `framer-motion` and changes text and button styles.
>   - **Code Refactoring**:
>     - Locks aspect ratio in `resize-handles.tsx` by setting `aspectRatioLocked` to `true`.
>     - Refactors variable names in `resize-handles.tsx` for clarity (e.g., `currentWidth` to `newWidth`).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 9f2affbb79c595bde58af6088bcb9aa27af1d5a6. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/web/client/src/components/store/editor/font/index.ts
+++ b/apps/web/client/src/components/store/editor/font/index.ts
@@ -1,4 +1,4 @@
-"use client";
+'use client';
 
 import type { ProjectManager } from '@/components/store/project/manager';
 import type { ParseResult } from '@babel/parser';
@@ -118,7 +118,9 @@ export class FontManager {
 
         // React to sandbox connection status
         const sandboxDisposer = reaction(
-            () => this.editorEngine.state.brandTab === BrandTabValue.FONTS && this.editorEngine.sandbox?.session.session,
+            () =>
+                this.editorEngine.state.brandTab === BrandTabValue.FONTS &&
+                this.editorEngine.sandbox?.session.session,
             (session) => {
                 if (session) {
                     this.loadInitialFonts();
@@ -128,7 +130,9 @@ export class FontManager {
         );
 
         const fontConfigDisposer = reaction(
-            () => this.editorEngine.state.brandTab === BrandTabValue.FONTS && this.editorEngine.sandbox?.readFile(this.fontConfigPath),
+            () =>
+                this.editorEngine.state.brandTab === BrandTabValue.FONTS &&
+                this.editorEngine.sandbox?.readFile(this.fontConfigPath),
             async (contentPromise) => {
                 if (contentPromise) {
                     const content = await contentPromise;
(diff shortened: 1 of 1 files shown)


kishansinghifs1:
@Kitenite  can you please review it .

Kitenite (changes requested):
Hey @kishansinghifs1 , 
Thanks for adding this. Seems like there are a lot of artifacts in the PR that should not be included. Could you clean up the PR first with only the relevant part? Thanks :)

Kitenite (changes requested):
This method is a regression, partial matching no longer works if there's a space which is not ideal. I think the ticket is not worth fixing but thanks for your time.

PR #2217, opened 2025-06-16
Fix top bar scale so external link does not hide

### LINKED ISSUE #2211: [bug] Topbar open link externally is hidden
#### Describe the bug

there's a button to open the link in a separate tab for a selected window, but for some reason it is only visible when you zoom-in far enough. This needs to be visible at any page size and the button should be anchored to the right side. 

Image

Also, let's swap the icon while we're in there: 

Image

As with all icons, make sure that it uses currentFill or the proper naming to be the right color.

JakeRoggenbuck:
## Description

Fix the scaling on the top bar so that the external link button does not get hidden when you zoom out too much.

## Related Issues

fixes #2211

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Testing

Zoomed far in:

Normal zoom:

Zoomed out:

## Additional Notes

----

> [!IMPORTANT]
> Fixes scaling issue in `TopBar` to ensure external link button remains visible when zoomed out.
> 
>   - **Behavior**:
>     - Fixes scaling issue in `TopBar` component in `top-bar.tsx` to prevent external link button from hiding when zoomed out.
>     - Introduces `shouldScale` logic to determine if scaling adjustments are needed based on `editorEngine.canvas.scale`.
>   - **UI Adjustments**:
>     - Adjusts `left` and `right` positioning and `transform` properties for elements within `TopBar` to maintain visibility at different zoom levels.
>     - Changes `overflow-hidden` to `overflow-visible` for the main `div` to ensure elements are not clipped.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 10274d031715f8627b7a46b2f129cf103ec351e9. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Kitenite (commented):
Is there not a flex solution for this instead of making everything absolute? Seems like the problem is that the input is not resizing correctly? This works pretty well so I'm inclined to merge it in but wonder if there's a sizing reason you found that flex grow or something wouldn't do?

JakeRoggenbuck:
Sounds good, I'll see if I can find a way to fix it with a flexbox class.

JakeRoggenbuck:
Small:



Normal:



Large Left:



Large Right:



When the window gets super small, the url does overlap with the external link button, but that's at like 11% size:

image

JakeRoggenbuck:
Yea, it should probably not display the top bar if the zoom is below 15%.

JakeRoggenbuck:
14% zoom:

image

17% zoom:

image

PR #2223, opened 2025-06-17
compress image before upload

spartan-vutrannguyen:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

### DIFF
--- a/apps/web/client/src/app/projects/import/local/_components/select-folder.tsx
+++ b/apps/web/client/src/app/projects/import/local/_components/select-folder.tsx
@@ -9,6 +9,7 @@ import { motion } from 'motion/react';
 import { useCallback, useRef, useState } from 'react';
 import { useProjectCreation } from '../_context/project-creation-context';
 import { StepContent, StepFooter, StepHeader } from './steps';
+import { compressImage } from '@onlook/utility/src/image';
 
 declare module 'react' {
     interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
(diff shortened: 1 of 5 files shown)


packages/utility/test/image.test.ts
```diff
@@ -0,0 +1,116 @@
+import { describe, expect, it, mock } from 'bun:test';
+import { compressImage, base64ToBlob } from '../src/image';
+
+// Mock browser-image-compression
+const mockCompressedFile = new File(['compressed'], 'compressed.jpg', { type: 'image/jpeg' });
+const mockImageCompression = mock(() => mockCompressedFile);
+const mockGetDataUrlFromFile = mock(() => 'data:image/jpeg;base64,compressed-data');
+
+// Mock the module
+mock.module('browser-image-compression', () => ({
```
Kitenite: Once you mock this dependency, there's no longer a point to this test. We wanted to make sure the dependency actually exhibit expected behavior and that the compressed image is legible. With this, we're testing nothing besides the fact that the dependency is called.

apps/web/client/src/app/projects/import/local/_components/select-folder.tsx
```diff
@@ -78,7 +73,36 @@ export const NewSelectFolder = () => {
...
 
             // Skip ignored files
-            if (IGNORED_FILES.includes(file.name)) {
+            if (IGNORED_UPLOAD_FILES.includes(file.name)) {
+                continue;
+            }
+
+            // Compress large images files
+            // But not svg
+            if (file.type.startsWith('image/') && file.type !== 'image/svg+xml') {
```
Kitenite, line 82: Are we sure SVG is the only type we don't handle? Seems like PNG is not even listed here...
https://www.npmjs.com/package/browser-image-compression

Perhaps we're better off compressing on the server side instead.

PR #2238, opened 2025-06-18
feat: asset management

spartan-vutrannguyen:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Introduces server-side image processing with `@onlook/image-server` using Sharp, supporting various image formats and integrating with existing systems.
> 
>   - **New Feature**:
>     - Introduces `@onlook/image-server` for server-side image processing using Sharp.
>     - Adds `compressImageServer` and `batchCompressImagesServer` functions in `compress.ts` for image compression.
>     - Supports JPEG, PNG, WebP, TIFF, GIF, BMP formats; skips ICO and SVG.
>   - **Integration**:
>     - Updates `package.json` to include `@onlook/image-server` and `sharp` as dependencies.
>     - Adds `imageRouter` to `api/root.ts` for image compression API.
>     - Modifies `file.ts` and `image.ts` in `utility` to support image file operations.
>   - **Testing**:
>     - Adds comprehensive tests in `image.test.ts` for image compression, format handling, and error cases.
>   - **Documentation**:
>     - Provides detailed usage instructions and API documentation in `README.md` for `@onlook/image-server`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for a20993bf0ea5c9bbf22e9e6d52f2d365f7a268f4. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


packages/utility/package.json
```diff
@@ -39,6 +39,7 @@
         "is-subdir": "^1.2.0",
         "mime-lite": "^1.0.3",
         "normalize-url": "^8.0.1",
+        "sharp": "^0.33.5",
```
Kitenite: This is exactly what we should not do. If sharp is accidentally imported into the browser or the preload script it will cause runtime error. I would err on the side of caution and create an explicitly server side package with image

packages/constants/src/files.ts
```diff
@@ -4,7 +4,7 @@ export const IGNORED_DIRECTORIES = [
     'node_modules',
     'dist',
     'build',
-    'public',
+    // 'public',
```
Kitenite: Given that IGNORED_DIRECTORIES is abiguously named, we will not know what effect this may have. Will we not read in any public folder into dev? 

I would suggest renaming this to be more specific for its current use case and create a new array specifically for why you're changing this

packages/utility/test/images/input/sea.jpg
Kitenite, line 1: should we be testing different file types instead of 3 jpg?

packages/utility/src/image.ts
```diff
@@ -1,27 +1,2 @@
...
-    const byteString = atob(base64.split(',')[1] ?? '');
-    const ab = new ArrayBuffer(byteString.length);
-    const ia = new Uint8Array(ab);
-    for (let i = 0; i < byteString.length; i++) {
-        ia[i] = byteString.charCodeAt(i);
-    }
-    return new Blob([ab], { type: mimeType });
-}
+export * from './image-types';
+export * from './image-client';
```
Kitenite: I see that the image-server file is not explicitly exported so maybe it's ok. It's still scary to have the sharp depedency in utility because preload will try to bundle it into the script. And if it uses node dependency it will crash the preload script. I think it should be in its own package still.

packages/utility/src/file.ts
```diff
@@ -97,3 +97,18 @@ export const getMimeType = (fileName: string): string => {
     if (res) return res;
     return 'application/octet-stream';
 };
+
+export const isImageFile = (fileName: string): boolean => {
+    const mimeType = getMimeType(fileName);
+    return ['image/jpeg', 'image/png', 'image/gif', 'image/webp', 'image/svg+xml'].includes(
```
Kitenite: I see a lot of hard coded mime types already. Can we centralize them in constants?

Kitenite: We also have a mime type library used somewhere but let's make sure they're not repeated everywhere if we can help it.

apps/web/client/src/components/store/editor/sandbox/index.ts
```diff
@@ -51,22 +54,28 @@ export class SandboxManager {
...
-                if (content === null) {
-                    console.error(`Failed to read file ${normalizedPath}`);
-                    continue;
-                }
+                const isImage = isImageFile(normalizedPath);
+
+                if (isImage) {
+                    // Only track image file path during indexing, don't read content
+                    this.fileSync.trackBinaryFile(normalizedPath);
+                } else {
```
spartan-vutrannguyen, line 65: I need indexing image so I move skip non-JSX files to here @Kitenite

packages/image-server/src/index.ts
```diff
@@ -0,0 +1,6 @@
+// ⚠️ WARNING: This package contains Node.js-only dependencies (Sharp)
+
+export * from './types';
+export * from './compress';
+
+export { compressImageServer, batchCompressImagesServer } from './compress';
```
Kitenite: is thre a reason for the extra export here given everything is already exported on line 4?

apps/web/client/src/app/projects/import/local/_components/select-folder.tsx
```diff
@@ -80,7 +94,7 @@ export const NewSelectFolder = () => {
 
             // Compress large images files
             // But not svg
-            if (file.type.startsWith('image/') && file.type !== 'image/svg+xml') {
+            if (file.type.startsWith('image/')) {
```
Kitenite, line 101: Hey @spartan-vutrannguyen , 
Could you help validate this change? I want to make sure we fix the original reason we moved to server side compression.

packages/image-server/test/images/input/dunes.png
Kitenite, line 1: these are likely too large to contain in the code. Could you add some very small images instead? We should probably test gif and webp if possible but I'm happy that these pass.

spartan-vutrannguyen: Sure, I added these large images to test the performance of this function. Since it works fine in local, i think now I can replace them

PR #2277, opened 2025-06-25
fix: devtab does not update after applying code changes

spartan-vutrannguyen:
## Description

- Problem: The Dev tab doesn't update after applying code changes.
- Root Cause: The event listener is cleared when the user switches to the Chat tab, so the Dev tab no longer receives updates.
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

### DIFF
--- a/apps/web/client/src/app/project/[id]/_components/right-panel/dev-tab/index.tsx
+++ b/apps/web/client/src/app/project/[id]/_components/right-panel/dev-tab/index.tsx
@@ -1,6 +1,7 @@
 import { useEditorEngine } from '@/components/store/editor';
 import type { CodeRange, EditorFile } from '@/components/store/editor/dev';
 import type { FileEvent } from '@/components/store/editor/sandbox/file-event-bus';
+import { useSubscriptionCleanup } from '@/hooks/use-subscription-cleanup';
 import { EditorView } from '@codemirror/view';
 import { Button } from '@onlook/ui/button';
 import {
@@ -26,6 +27,7 @@ import { FolderModal } from './folder-modal';
 export const DevTab = observer(() => {
     const editorEngine = useEditorEngine();
     const { theme } = useTheme();
+    const { addSubscription } = useSubscriptionCleanup();
     const ide = editorEngine.ide;
     const [isFilesVisible, setIsFilesVisible] = useState(true);
     const [fileModalOpen, setFileModalOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


apps/web/client/src/app/project/[id]/_components/right-panel/dev-tab/index.tsx
```diff
@@ -209,13 +211,13 @@ export const DevTab = observer(() => {
...
-        // Subscribe to all file events
+
         const unsubscribe = editorEngine.sandbox.fileEventBus.subscribe('*', handleFileEvent);
 
-        return () => {
-            unsubscribe();
-        };
-    }, [editorEngine.sandbox, ide.activeFile]);
+        // Use the subscription cleanup hook
+        addSubscription(unsubscribe);
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/right-panel/dev-tab/index.tsx]: This will add the subscription multiple time on re-render and create memory leak. We need to dedup or something. 

But taking a step back on this implementation. This is applying a complex solution onto an issue introduced by complexity instead of removing the complexity. 

Why is cleaning this on un-render causing weird behavior? The listeners should not be persistent for the lifetime of the app. If the component is un-rendered, it should stop listening to events. Why is it even subscribing and not using MobX to listen to state changes? Why is there even an eventbus when we already have a subscription pattern with MobX?

Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/right-panel/dev-tab/index.tsx]: The quick fix here is to add a key and dedup. But I think the better solution is to have the devTab listen directly to the files in the sandbox files manager. This remove the file watcher and remove the eventBus.

apps/web/client/src/hooks/use-subscription-cleanup.ts
```diff
@@ -0,0 +1,62 @@
...
+import { useCallback, useEffect, useRef } from 'react';
+
+type UnsubscribeFunction = () => void;
+
+/**
+ * Custom hook to manage subscriptions with automatic cleanup on navigation changes
+ * WARNING: This hook only cleans up subscriptions when the pathname changes.
+ * @returns Object with addSubscription function and manual cleanup function
+ */
+export const useSubscriptionCleanup = () => {
```
Kitenite: This naming needs to be better. It doesn't tell us anything important. Ideally, the fact that this unsubscribe when page changes needs to be evident in the function name such that consumers of this does not need to read comments/documentation to know how to use it.

Kitenite: Something like `useCleanupOnPageChange`

PR #2282, opened 2025-06-26
Fix/update overlay in resizing

### LINKED ISSUE #2278: [bug] The selected overlay should be updated when resizing windows
#### Describe the bug

Image

gautamtayal1:
## Description

- Added boundary clipping so overlays get cut off at iframe edges
- Added tests

## Related Issues

fixes #2278 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Testing

iframe-bounds.test.ts

Screenshot 2025-06-26 at 8 03 13 PM

https://github.com/user-attachments/assets/adb1aa66-0d9f-42d7-9d6b-a015925f0f83


Kitenite (changes requested):
The resize change works really well! 

Unfortunately, we should not be clipping the rectangles on the boundary of the frame. I would hate to waste the good work there but it's actually expected behavior to go out of frame (Framer example below). 

Could you only keep the resize update change in this PR and I can merge that in?

image
image

gautamtayal1:
@Kitenite, I've removed all the clipping logic and kept just the resize updates. Sorry for implementing the clipping logic in the first place! I misunderstood the expected behavior. Thanks for the feedback.

apps/web/client/src/components/store/editor/overlay/index.ts
```diff
@@ -7,6 +7,8 @@ import { adaptRectToCanvas } from './utils';
 
 export class OverlayManager {
     state: OverlayState = new OverlayState();
+    private resizeObserver: ResizeObserver | null = null;
```
Kitenite: Instead of an explicit resize observer on the iframe, what do you think about calling refresh on the overlay from resize-handler instead on mouse-move? Seems like it would save us the extra complexity and memory overhead of an observer?

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
```diff
@@ -67,12 +69,16 @@ export const ResizeHandles = observer(({ frame }: { frame: FrameImpl }) => {
                 width: Math.round(newWidth),
                 height: Math.round(newHeight),
             };
+
+            editorEngine.overlay.undebouncedRefresh();
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx:71]: Seems like you're already calling refresh here which makes sense, then should we even be using the observer?

Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx:71]: How bad is the debounced version here instead of un-debounced?

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
```diff
@@ -26,6 +26,8 @@ export const ResizeHandles = observer(({ frame }: { frame: FrameImpl }) => {
         const startHeight = frame.dimension.height;
         const aspectRatio = startWidth / startHeight;
 
+        editorEngine.state.iframeResizing = true;
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx]: If we're just calling refresh, is there a point to this flag?

apps/web/client/src/components/store/editor/state/index.ts
```diff
@@ -15,6 +15,8 @@ export class StateManager {
     publishOpen = false;
     leftPanelLocked = false;
     canvasPanning = false;
+    plansOpen = false;
```
Kitenite: Not sure where this came from

apps/web/client/tsconfig.json
```diff
@@ -17,6 +17,7 @@
     },
     "include": [
         "src",
+        "test",
```
Kitenite: What is this for?

Kitenite:
> @Kitenite, I've removed all the clipping logic and kept just the resize updates. Sorry for implementing the clipping logic in the first place! I misunderstood the expected behavior. Thanks for the feedback.

No worries! I like the initiative on it.

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx]: Ah I think it's from the original clipping change. Makes sense but we should remove this?

apps/web/client/src/components/store/editor/overlay/index.ts
gautamtayal1: you are right. removed this part.

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
Comment (gautamtayal1) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx:71]: i tested the debounced version but it introduced noticeable lag.

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
Comment (gautamtayal1) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx]: yes, this part is unnecessary now, my bad. removed this.

apps/web/client/src/components/store/editor/state/index.ts
gautamtayal1: this was already a part of codebase, I removed it during a merge conflict by mistake. So re-added it.

apps/web/client/tsconfig.json
gautamtayal1: I was facing some ts errors on imports when I made test which I removed with the boundary logic. So, I'm removing this now.

Kitenite: Yeah the tsconfig is (ironically) always throwing weird linter errors

apps/web/client/src/components/store/editor/state/index.ts
Kitenite: I think it was since removed so I think we can remove it.

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx:71]: Yep that makes sense thanks for checkin that. I think this is uncommon enough that it's ok to take the performance hit.

apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx
Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/resize-handles.tsx]: awesome thanks @gautamtayal1 !

gautamtayal1:
@Kitenite, I’ve implemented the requested changes. Thanks for your patience and feedback - I learned a lot.
Let me know if there’s anything else you’d like me to tweak.

apps/web/client/src/components/store/editor/overlay/index.ts
```diff
@@ -39,7 +41,7 @@ export class OverlayManager {
                 console.error('Element not found');
                 continue;
             }
-            const adaptedRect = adaptRectToCanvas(el.rect, view);
+            const adaptedRect = adaptRectToCanvas(el.rect, view, false);
```
Kitenite: Is there a reason this should be changed?

Kitenite: Reverting since it's out of scope

PR #2289, opened 2025-06-27
feat: version manager

spartan-vutrannguyen:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Introduces version management with Git integration, enabling commit creation, listing, and restoration, along with UI and database updates to support these features.
> 
>   - **Behavior**:
>     - Introduces version management using Git in `git-manager.ts`, allowing commit creation, listing, and restoration.
>     - Updates `ChatManager` in `chat/index.ts` to create commits when sending messages.
>     - Adds commit OID handling in `UserChatMessageImpl` in `chat/message/user.ts`.
>     - Implements commit restoration in `UserMessage` in `chat-messages/user-message.tsx`.
>   - **UI Components**:
>     - Adds version management UI in `versions.tsx`, `version-row.tsx`, and related components.
>     - Displays saved and available versions, allowing users to create, rename, and restore versions.
>   - **Database**:
>     - Updates `messages` schema in `message.ts` to include `commitOid` field.
>     - Modifies `toMessage` and `fromMessage` in `dto/message.ts` to handle `commitOid`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 92c11eaef463c8035fe6e814cf7c537cbe20b5c3. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/src/components/store/project/git-manager.ts
```diff
@@ -0,0 +1,329 @@
...
+                    timestamp: timestamp,
+                    displayName: message || null,
+                });
+            }
+        }
+
+        return commits;
+    }
+
+    private formatGitLogOutput(input: string): string {
```
Kitenite, line 293: Why not use the existing library instead?

apps/web/client/src/components/store/project/git-manager.ts
```diff
@@ -0,0 +1,329 @@
...
+
+            let result = await this.editorEngine.sandbox.session.runCommand(command);
+
+            // If the command failed due to shell not being active, try to reconnect and retry once
+            if (!result.success && result.error?.includes('Shell with id') && result.error?.includes('is not active')) {
+                console.log('Shell not active, attempting to ensure session is ready and retry...');
+
+                // Try to ensure session is ready
+                const sessionManager = this.editorEngine.sandbox.session;
+                if ('ensureSessionReady' in sessionManager && typeof sessionManager.ensureSessionReady === 'function') {
```
Kitenite: What is the point of this check? It's already type-safe

Kitenite: I don't know what happened here but entire code block is not up to your usual standard.

1. The type checkfor the function existing on this makes no sense. 
2. You created a sessionManager constant for no good reason when it's being used on top without the constant. 
3. You created a function to check terminal session when we are not using terminal session to run this command. 
4. You created a redundant session active check when we already have one. (called ping)
5. You make unsafe string checks on the error that can and very likely will change over time. 

I think in this case it's better just to throw the error and retry than do this.

apps/web/client/src/components/store/project/git-manager.ts
```diff
@@ -0,0 +1,329 @@
...
+                    if (isReady) {
+                        console.log('Session ready, retrying command...');
+                        result = await this.editorEngine.sandbox.session.runCommand(command, (output) => {
+                            console.log(`${command} retry output:`, output);
+                        });
+                    }
+                }
+            }
+
+            return {
```
Kitenite: Why not just return the entire object here? It's the exact same object

PR #2302, opened 2025-06-28
fix: remove managers in favor of queries

Kitenite:
## Description

- Remove user manager
- Remove projects manager
- Make editorEngine scoped to page with cleanup
- Update user table

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [x] Refactor

## Additional Notes

----

> [!IMPORTANT]
> This PR removes `userManager`, replacing it with direct `trpc` API calls and introduces `StateManager` for UI state management, affecting user and project management components.
> 
>   - **State Management**:
>     - Introduces `StateManager` in `state/manager.ts` to manage UI state like `isSubscriptionModalOpen` and `isSettingsModalOpen`.
>     - Replaces `userManager` with `StateManager` in components like `avatar-dropdown/index.tsx` and `pricing-modal.tsx`.
>   - **API Changes**:
>     - Replaces `userManager` with direct `trpc` API calls in components for user data and settings.
>     - Updates `userRouter` in `user/user.ts` to handle user data and settings directly from the database.
>     - Modifies `projectRouter` in `project/project.ts` to manage projects directly via `trpc`.
>   - **Component Updates**:
>     - Refactors components like `settings-modal/with-project.tsx` and `settings-modal/without-project.tsx` to use `StateManager` for modal state.
>     - Updates `ProjectTab`, `SiteTab`, and `PageTab` to use `trpc` for project and page settings.
>     - Adjusts `VersionRow` and `NoVersions` components to use `StateManager` for UI interactions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 37070dbe0987ae9cae5c37dc36287c7337a3b9a2. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/src/app/projects/_components/select.tsx
```diff
@@ -9,16 +9,15 @@ import { Carousel } from './carousel';
 import { ProjectInfo } from './info';
 
 export const SelectProject = observer(() => {
-    const projectsManager = useProjectsManager();
+    const { data: projects, isLoading } = api.project.list.useQuery();
```
http-teapot, line 12: Unless I am misreading, I think this pulls all projects? In other words, if a user creates a 1000 projects, it'll pull all of them?

apps/web/client/src/app/projects/_components/select.tsx
```diff
@@ -9,16 +9,15 @@ import { Carousel } from './carousel';
...
-        .toSorted(
-            (a, b) =>
-                new Date(b.metadata.updatedAt).getTime() -
-                new Date(a.metadata.updatedAt).getTime(),
-        );
+    const sortedProjects = projects?.toSorted(
+        (a, b) =>
+            new Date(b.metadata.updatedAt).getTime() -
+            new Date(a.metadata.updatedAt).getTime(),
+    ) ?? [];
```
http-teapot, line 20: Could this be done at the query level? That way you can sort and limit the query; much more performant.

apps/web/client/src/app/projects/import/github/_context/context.tsx
```diff
@@ -136,7 +135,7 @@ export const ImportGithubProjectProvider: React.FC<ImportGithubProjectProviderPr
 
     // Create manager
     const createManager = useCreateManager();
-    const userManager = useUserManager();
+    const { data: user } = api.user.get.useQuery();
```
http-teapot, line 138: I am not super familiar with the code but reviewing this on the underlying code:

```ts
        const authUser = ctx.user;
        const user = await ctx.db.query.users.findFirst({
            where: eq(users.id, authUser.id),
        });
```
What happens if the function is called `user` is not defined on `ctx` or the code instantiating the property hasn't ran and the `user` object is not loaded?

apps/web/client/src/components/store/editor/chat/conversation/index.ts
```diff
@@ -14,8 +13,7 @@ export class ConversationManager {
     private _conversations: ChatConversation[] = [];
 
     constructor(
-        private chatManager: ChatManager,
-        private projectManager: ProjectManager,
+        private projectId: string,
```
http-teapot, line 16: Top of my mind at this moment: how do you guarantee that parameters like this one is set and is correct?

Examples:
1. If I put an invalid `uuid` in the URL
2. If I put a valid `uuid` in the URL but it doesn't exist
3. If I switch between projects – how is the state handled, is there any race/timing conditions

apps/web/client/src/components/ui/settings-modal/project/index.tsx
```diff
@@ -18,36 +23,51 @@ export const ProjectTab = observer(() => {
...
     const nameInput = useDebouncedInput(
         name,
-        (value) => projectsManager.updatePartialProject({ name: value })
+        (value) => updateProject({
+            id: editorEngine.projectId,
+            project: {
+                ...project,
+                name: value,
+            },
+        })
```
http-teapot, line 32: This pattern might cause you some headaches in the future. I am not super familiar with Supabase and the libraries in between but, if `project` doesn't have all its properties (for example: the code only queries for the `name`) then it may reset the missing properties to their default values (`null` or error if non nullable).

I'd recommend updating only the column `name` and not the entire object.

Though it technically works now, it's one of those things that may slow you down later. Quite annoying to refactor, much easier to use proper pattern now.

However, like I said, I didn't verify this and I am just reflecting on past experiences.

apps/web/client/src/components/ui/settings-modal/site/page.tsx
```diff
@@ -8,11 +8,11 @@ import { MetadataForm } from './metadata-form';
...
 
 export const PageTab = memo(({ metadata, path }: { metadata?: PageMetadata; path: string }) => {
-    const projectsManager = useProjectManager();
-    const domainsManager = useDomainsManager()
     const editorEngine = useEditorEngine();
-    const project = projectsManager.project;
-    const baseUrl = domainsManager.domains.preview?.url ?? domainsManager.domains.preview?.url ?? project?.sandbox.url;
+    const { data: project } = api.project.get.useQuery({ projectId: editorEngine.projectId });
+    const { data: domains } = api.domain.getAll.useQuery({ projectId: editorEngine.projectId });
+    const baseUrl = domains?.published?.url ?? domains?.preview?.url ?? project?.sandbox?.url;
```
http-teapot, line 14: Another part I am not familiar with in Next.js: how does reactivity work here?

`useQuery` may trigger I/O, it's my understanding that there is a property `isLoading` along with `data` so I assume that `data` can contain different values and updates over time.

Shouldn't there be a reactive block on line 14 so it reacts to changes on variables `project` and `domains`?

package.json
```diff
@@ -19,8 +19,8 @@
     "scripts": {
         "build": "bun --filter @onlook/web build:client",
         "start": "bun --filter @onlook/web start:client",
-        "dev": "bun --filter @onlook/web dev",
-        "test": "bun --filter '*' test",
+        "dev": "bun --elide-lines 0 --filter @onlook/web dev",
```
http-teapot, line 22: shocked that `bun dev -- --elide-lines 0` didn't work but glad this works

apps/web/client/src/app/projects/_components/select.tsx
Kitenite: Yeah we can probably paginate but the project shell themselves are pretty small so I think it's ok for now. Average user should not get over 20 projects in theory?

apps/web/client/src/app/projects/_components/select.tsx
Kitenite: That's true, will update this

package.json
Kitenite: weirdly adding the flag at the end doesn work, it had to be before `--filter` for some reason

apps/web/client/src/components/ui/settings-modal/site/page.tsx
Kitenite: data can be undefined until the query resolves. 
isLoading is just a more ensured loading state since it could be undefined for other reason than not yet loaded.

Kitenite: 14 will update because data changing will trigger a rerender. Doesn't really need an extra useEffect there.

apps/web/client/src/components/ui/settings-modal/project/index.tsx
Kitenite: Yeah this is actually really bad, we should only need to feed in the values that matter. Good catch here, we should not spread the project in.

apps/web/client/src/components/store/editor/chat/conversation/index.ts
Kitenite: Good point, it'll break but we should do an existence check. We'll also do a check later once we have ownership implemented to make sure user should be accessing the project.

apps/web/client/src/app/projects/import/github/_context/context.tsx
Kitenite: When the function is defined, it can be set as a `protectedProcedure` or a `publicProcedure`. Protected procedure will throw an UNAUTHORIZED if a user context is not found.

apps/web/client/src/app/projects/_components/select.tsx
http-teapot: In my past project, I sorta had the same approach but I ended up piling on a bunch of vulnerabilities (from a productivity standpoint). We'd get spike traffic (like hundred users loading the same page as the same time, so not that much) and it ended up taking down the database. It just took a while to locate the code responsible for it and stopped us from iterating.

The other thing I just thought of, in our SQL query logs, there is no way to identify where that query was being executed from. We'd use GraphQL and our GraphQL queries were named but those names didn't show the logs.

Not for now but might be worth finding a system to associate a query and the location it's executed from.

PR #2364, opened 2025-07-07
Fix: Directory picker not working when uploading local projects

AissaSemaoui:
## Description

There is a bug that prevents selecting directories when trying to upload a local project. 

I removed the accepts attributes as it accepts only files, and the webkitdirectory attribute allows only selecting directories, so this conflict ends up not being able to select anything. 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

Before : 

https://github.com/user-attachments/assets/1ea56326-abbf-4a39-983b-1c4995f74e04

After : 

https://github.com/user-attachments/assets/c823d9f9-b150-4115-b9b3-a6783714a23a


Kitenite (approved):
Huh, this may be an issue for Linux specifically. Works in me testing. Thanks for the fix @AissaSemaoui !

PR #2381, opened 2025-07-08
fix: height change on page reload

### LINKED ISSUE #2368: [bug] Doesn't seem to write unit changes of width / height to code
#### Describe the bug

I tried adjusting this height from 8 PX to 8 REM but after selecting the dropdown, then making another edit, it reverts back to 8 PX on page reload.

Image

Image
Image

gautamtayal1:
## Description

fixes font change on page reload. 

## Related Issues

closes #2368

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Fixes font change on page reload by converting CSS property values to lowercase in `createCSSRuleString()` in `tailwind.ts`.
> 
>   - **Bug Fix**:
>     - Fixes font change on page reload by converting CSS property values to lowercase in `createCSSRuleString()` in `tailwind.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 7113ed0f533952bd9b2c3c616d326566f540ac01. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/web/client/src/app/project/[id]/_components/editor-bar/inputs/input-dropdown.tsx
+++ b/apps/web/client/src/app/project/[id]/_components/editor-bar/inputs/input-dropdown.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { UNITS } from '@onlook/constants';
 import { Button } from '@onlook/ui/button';
 import {
     DropdownMenu,
@@ -9,8 +10,6 @@ import {
 } from '@onlook/ui/dropdown-menu';
 import { Icons } from '@onlook/ui/icons';
 import { useInputControl } from '../hooks/use-input-control';
-
-const UNITS = ['px', '%', 'em', 'rem'];
 
 const OPTION_OVERRIDES: Record<string, string | undefined> = {
     Fit: 'Hug',
--- a/apps/web/client/src/app/project/[id]/_components/editor-bar/inputs/input-dropdown.tsx
+++ b/apps/web/client/src/app/project/[id]/_components/editor-bar/inputs/input-dropdown.tsx
@@ -10,7 +10,7 @@ import {
 import { Icons } from '@onlook/ui/icons';
 import { useInputControl } from '../hooks/use-input-control';
 
-const UNITS = ['PX', '%', 'EM', 'REM'];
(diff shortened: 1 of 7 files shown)


Kitenite:
@gautamtayal1 , I couldn't reproduce this bug, could you share a before and after behavior if possible? 
and reason for the lowercase change?

gautamtayal1:
When we try to change the unit of let's say height, from px to rem, the height change renders on ui but is not persisted in code because we are sending uppercase unit from frontend (px -> PX) which creates a conflict and tailwind classname does not update. This change makes sure that the unit is lowercase. cc @Kitenite 

Behaviour (height):

Before any change - 
Screenshot 2025-07-08 at 11 34 39 PM

Without lowercase - 
Screenshot 2025-07-08 at 11 34 39 PM

With lowercase - 
Screenshot 2025-07-08 at 11 34 25 PM

Kitenite:
That makes sense. Any concern with case sensitive values turning lower case? 
This seems like a low-level spot to do this. We could do it when we write the unit or perhaps check for unit there before writing? 

But this makes sense, thanks for clarifying @gautamtayal1

gautamtayal1:
We can make the change directly in the units array:

const UNITS = ['PX', '%', 'EM', 'REM'] -> const UNITS = ['px, '%', 'em', 'rem']

this would be the easiest fix but we won't see the units in capital in the dropdown menu. Other option is to use tolowercase everywhere the units array is used in react components because units are used in multiple place.

gautamtayal1:
@Kitenite let me know which one you prefer, or if you have any other idea around it.

Kitenite:
I think we should keep the units the way they should be sent into CSS (lowercase) and upper case them when need to display.

gautamtayal1:
great suggestion, @Kitenite. I have made the changes you suggested. Let me know if I missed something.

Kitenite (approved):
Awesome thanks @gautamtayal1 !

PR #2426, opened 2025-07-11
Refactor subscription backend logic

http-teapot:
## Description

1. Updated the subscription lifecycle to use Stripe webhooks only.
2. Added rate limiting logic tied to subscription lifecycle
3. Updated frontend to use rate limiting when user has active subscription

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [x] Refactor

## Testing

1. Make sure you start with a clean DB.
2. Log in.
3. Click the profile dropdown, make sure it reflects the free rates.
4. Create a new project or update an existing project (interact with the AI).
5. Verify that the usage counted towards the free rate limit.
6. Purchase a new subscription.
7. Verify that the limit reflects the purchased subscription and that the previous message does not count towards the limit.
8. Post a message (similar as step 4).
9. Verify that the usage counted towards the subscription monthly limit.
10. Upgrade the subscription to the next tier.
11. Verify in the dropdown that the rate limit as increased accordingly to the delta between the previous pricing and the new pricing.
12. Post a message
13. Verify that the usage has increased accordingly.
14. Simulate a renewal
     1. Go to Stripe Sandbox
     2. Find and click on the subscription
     3. In the Actions dropdown, click "Update subscription"
     4. Add a new tier (add product) and remove the existing tier (Stripe handles that as an upgrade)
     5. Check "Reset billing cycle" (this will invoice the user, just like it'd if it renewed)
     6. Click "Update subscription"
15. The new rate limit should include the previous rate limit minus the used credits
16. Repeat step (14)
17. The new rate limit should include the previous rate limit but the one before that (ie. credits are carried over only once)

## Additional Notes

----

> [!IMPORTANT]
> Refactor subscription backend to use Stripe webhooks and implement rate limiting based on subscription status, updating database schema and frontend accordingly.
> 
>   - **Behavior**:
>     - Refactor subscription lifecycle to use Stripe webhooks exclusively in `route.ts` and `stripe.ts`.
>     - Implement rate limiting logic tied to subscription lifecycle in `route.ts` and `usage/index.ts`.
>     - Update frontend to reflect rate limiting based on subscription status in `route.ts`.
>   - **Database**:
>     - Add `rateLimits` table in `rate-limits.ts` to track usage limits per subscription.
>     - Update `subscriptions` table in `subscription.ts` to include period start and end timestamps.
>     - Modify `usageRecords` table in `usage.ts` to include indexing for efficient querying.
>   - **Stripe Integration**:
>     - Add `createCustomer` function in `functions.ts` to handle new customer creation.
>     - Modify `createCheckoutSession` to include `stripeCustomerId` in `functions.ts`.
>     - Update subscription handling functions to manage subscription updates and cancellations in `functions.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for bf4cdb46e0a832643c3dd5e697ff703e12016cb5. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/src/app/webhook/stripe/stripe.ts
```diff
@@ -1,186 +1,323 @@
...
+
+                // Here you can decide the logic for the carry-over.
+                // Example, you may want to carry over 100% of the credits on the first carry-over,
+                // and 50% of the credits on the next carry-overs.
+                // const max = rate.carryOverTotal === 0 ? rate.left : rate.left * 0.50;
+                const max = rate.left;
+
+                // For now, we only carry over the credits on the first carry-over.
+                // In the future, we may want to carry over the credits on the next carry-overs.
+                if (rate.carryOverTotal === 0) {
```
http-teapot: This is where you can adjust the carry-over rules. Right now it is set up as:
1. It only carries over once.
2. 100% of the remaining credits carry over.

apps/web/client/src/app/api/chat/route.ts
```diff
@@ -143,25 +161,16 @@ export const streamResponse = async (req: NextRequest) => {
 
             return { ...toolCall, args: JSON.stringify(repairedArgs) };
         },
-        onError: (error) => {
+        onError: async (error) => {
             console.error('Error in chat', error);
+            // if there was an error with the API, do not penalize the user
+            if (usageRecordId && rateLimitId) {
```
Kitenite, line 184: This makes a lot of sense. Thanks for implementing the revert

apps/web/client/src/app/webhook/stripe/stripe.ts
```diff
@@ -1,186 +1,323 @@
...
+                stripeSubscriptionId: stripeSubscriptionId,
+                stripeSubscriptionItemId: stripeSubscriptionItemId,
+                stripeCurrentPeriodStart: currentPeriodStart,
+                stripeCurrentPeriodEnd: currentPeriodEnd,
+            })
+            .onConflictDoUpdate({
+                target: [subscriptions.stripeSubscriptionItemId],
+                set: {
+                    // Left in case there are concurrent webhook requests for the same subscription
+                    status: 'active',
```
Kitenite: Would this inadvertently mark a cancelled subscription active? 
If this is intended, should we also update the subscription period/other metadata?

apps/web/client/src/app/webhook/stripe/stripe.ts
```diff
@@ -1,186 +1,323 @@
...
+    receivedEvent: Stripe.CustomerSubscriptionDeletedEvent,
+) => {
+    const { stripeSubscriptionId } = extractIdsFromEvent(receivedEvent);
+
+    const res = await db
+        .update(subscriptions)
+        .set({
+            status: 'canceled',
+            endedAt: new Date(),
+            scheduledPriceId: null,
```
Kitenite: Is there a reason we would not want to keep these values for recording keeping purposes?

apps/web/client/src/app/webhook/stripe/stripe.ts
```diff
@@ -1,186 +1,323 @@
...
+                    stripeSubscriptionItemId,
+                });
+            } else {
+                // If it is not pro-rated, then it is a completely new period.
+                // Therefore, it should behave similarly to a renewal: credits need to be carried over.
+                renew = true;
+            }
+        });
+    } else if (
+        stripeSubscription.status === 'active' &&
```
Kitenite: So this block signifies updating the payment period changing? As in a renew? Could help to name this case

```
const  isStartingNewPeriod = stripeSubscription.status === 'active' &&
        +currentPeriodEnd !== +subscription.stripeCurrentPeriodEnd
```

apps/web/client/src/server/api/routers/subscription/subscription.ts
```diff
@@ -51,9 +51,38 @@ export const subscriptionRouter = createTRPCRouter({
...
+        let stripeCustomerId = userData?.stripeCustomerId;
+        if (!stripeCustomerId) {
+            // Store Stripe's customer ID as it is available in all customer-related events and
+            // API requests.
+            // Important, it may seem like a good idea to check if the customer already exists
+            // by looking up the email in Stripe, however, this can be a security risk since
+            // a user may sign up with an email that is not their own.
+            // This may happen when a user changes their email address in the app and the email
+            // is not updated in Stripe.
+            const customer = await createCustomer({
```
Kitenite, line 83: Should we actually add the user ID into the stripe checkout session and add the custom ID on webhook instead? This way we let stripe handle creating the customer instead of creating them manually?

apps/web/client/src/server/api/routers/subscription/subscription.ts
```diff
@@ -118,34 +147,17 @@ export const subscriptionRouter = createTRPCRouter({
         const isUpgrade = newPrice?.monthlyMessageLimit > currentPrice.monthlyMessageLimit;
```
Kitenite: This is a now repeat logic. Perhaps it makes sense to centralize this business logic for determining price upgrade

apps/web/client/src/server/api/routers/subscription/subscription.ts
Kitenite: Note: deciding this is better

apps/web/client/src/server/api/routers/usage/index.ts
```diff
@@ -77,10 +99,58 @@ export const usageRouter = createTRPCRouter({
         type: z.nativeEnum(UsageType),
     })).mutation(async ({ ctx, input }) => {
         const user = ctx.user;
-        await db.insert(usageRecords).values({
-            userId: user.id,
-            type: input.type,
-            timestamp: new Date(),
+        // running a transaction helps with concurrency issues and ensures that
+        // the usage is incremented atomically
+        return db.transaction(async (tx) => {
```
Kitenite: What happens when increment is called on a free plan with no rateLimits? Seems like in this case there should be an upsert or some logic here to handle that? Seeing a 500 error when running with an existing user with no subscription

Kitenite:
```
 @onlook/web-client dev: [TRPC] usage.increment took 327ms to execute
│ @onlook/web-client dev:  ⨯ Error [TRPCError]: Rollback
│ @onlook/web-client dev:     at <unknown> (src/server/api/routers/usage/index.ts:122:19)
│ @onlook/web-client dev:     at async (src/server/api/trpc.ts:105:19)
│ @onlook/web-client dev:     at async streamResponse (src/app/api/chat/route.ts:102:29)
│ @onlook/web-client dev:   120 |             // if there are no credits left then rollback
│ @onlook/web-client dev:   121 |             if (!limit?.left) {
│ @onlook/web-client dev: > 122 |                 tx.rollback();
│ @onlook/web-client dev:       |                   ^
│ @onlook/web-client dev:   123 |                 return;
│ @onlook/web-client dev:   124 |             }
│ @onlook/web-client dev:   125 | {
│ @onlook/web-client dev:   code: 'INTERNAL_SERVER_ERROR',
│ @onlook/web-client dev:   [cause]: Error [DrizzleError]: Rollback
│ @onlook/web-client dev:       at <unknown> (src/server/api/routers/usage/index.ts:122:19)
│ @onlook/web-client dev:       at async (src/server/api/trpc.ts:105:19)
│ @onlook/web-client dev:       at async streamResponse (src/app/api/chat/route.ts:102:29)
│ @onlook/web-client dev:     120 |             // if there are no credits left then rollback
│ @onlook/web-client dev:     121 |             if (!limit?.left) {
│ @onlook/web-client dev:   > 122 |                 tx.rollback();
│ @onlook/web-client dev:         |                   ^
│ @onlook/web-client dev:     123 |                 return;
│ @onlook/web-client dev:     124 |             }
│ @onlook/web-client dev:     125 | {
│ @onlook/web-client dev:     cause: undefined
│ @onlook/web-client dev:   }
│ @onlook/web-client dev: }
│ @onlook/web-client dev:  ⨯ Error [TRPCError]: Rollback
│ @onlook/web-client dev:     at <unknown> (src/server/api/routers/usage/index.ts:122:19)
│ @onlook/web-client dev:     at async (src/server/api/trpc.ts:105:19)
│ @onlook/web-client dev:     at async streamResponse (src/app/api/chat/route.ts:102:29)
│ @onlook/web-client dev:   120 |             // if there are no credits left then rollback
│ @onlook/web-client dev:   121 |             if (!limit?.left) {
│ @onlook/web-client dev: > 122 |                 tx.rollback();
│ @onlook/web-client dev:       |                   ^
│ @onlook/web-client dev:   123 |                 return;
│ @onlook/web-client dev:   124 |             }
│ @onlook/web-client dev:   125 | {
│ @onlook/web-client dev:   code: 'INTERNAL_SERVER_ERROR',
│ @onlook/web-client dev:   [cause]: Error [DrizzleError]: Rollback
│ @onlook/web-client dev:       at <unknown> (src/server/api/routers/usage/index.ts:122:19)
│ @onlook/web-client dev:       at async (src/server/api/trpc.ts:105:19)
│ @onlook/web-client dev:       at async streamResponse (src/app/api/chat/route.ts:102:29)
│ @onlook/web-client dev:     120 |             // if there are no credits left then rollback
│ @onlook/web-client dev:     121 |             if (!limit?.left) {
│ @onlook/web-client dev:   > 122 |                 tx.rollback();
│ @onlook/web-client dev:         |                   ^
│ @onlook/web-client dev:     123 |                 return;
│ @onlook/web-client dev:     124 |             }
│ @onlook/web-client dev:     125 | {
│ @onlook/web-client dev:     cause: undefined
│ @onlook/web-client dev:   }
│ @onlook/web-client dev: }
```

apps/web/client/src/server/api/routers/usage/index.ts
```diff
@@ -77,10 +98,67 @@ export const usageRouter = createTRPCRouter({
...
+        // running a transaction helps with concurrency issues and ensures that
+        // the usage is incremented atomically
+        return ctx.db.transaction(async (tx) => {
+            // users on free plans don't have their rate limits stored in the database
+            // the limits are calculated on the fly instead
+            const subscription = await tx.query.subscriptions.findFirst({
+                where: and(eq(subscriptions.userId, user.id), eq(subscriptions.status, SubscriptionStatus.ACTIVE)),
+            });
+            if (!subscription) {
+                return { rateLimitId: undefined, usageRecordId: undefined };
```
Kitenite: I think in this case we need to check the records for the free user to make sure they don't go over their free tier limit. Would it make sense to create a rate-limit row? Or just use our custom logic and count the usage record?

This would also not count towards free usage limit because it's a no-op. Ideally we still want to make sure this works. You can simulate free user by editing the subscription table row and marking the subscription as cancelled. 
Screenshot 2025-07-18 at 12 47 23 PM

http-teapot: Ah yeah, great catch

I didn't want to have rows for free users since they have daily rates, if you have a million users then each month would add 30 million rows in the table. On the other hand, your comment does highlight an issue with separating systems and for free users, it involves counting rows on a large table (though, now indexed) which is not optimal.

I was trying to keep it simple so the credits stored in the database were tied to the subscription's lifecycle. The alternative would be to inject two rate limits (one daily and one monthly) when a free user sends a message and then when a free user converts to paid, have the daily/monthly rate limits `endedAt` in the past.

IMO, it's the sort of thing you want to stabilize so it doesn't have to be revisited until months/years from now. By that time, you may revisit how you monetize things and start from scratch.

Kitenite: Yeah I see the problem with the usage calculation now. Especially when scaled up. Perhaps having a daily usage count system could make sense. 

One last wrinkle here is that the daily usage counting is now only monthly limit counting when user doesn't have a subscription. Seems like this would not reset daily so free users will only have 5 free messages per month. Please let me know if this is accurate. 

https://github.com/onlook-dev/onlook/blob/b5305af9aeb4bf556be19ce3da00a0374cc4b160/apps/web/client/src/server/api/routers/usage/index.ts#L82

packages/stripe/src/functions.ts
```diff
@@ -166,4 +172,13 @@ export const releaseSubscriptionSchedule = async ({
...
+};
+
+export const getSubscriptionSchedule = async ({
+    subscriptionScheduleId,
+}: {
+    subscriptionScheduleId: string;
+}) => {
+    const stripe = createStripeClient();
+    return stripe.subscriptionSchedules.retrieve(subscriptionScheduleId);
+};
```
http-teapot, line 184: Changes are here

packages/db/src/seed/stripe/backfill-subscriptions.ts
```diff
@@ -0,0 +1,121 @@
...
+        endedAt: endDate,
+        max: item.subscription.price.monthlyMessageLimit,
+        left: remainingUsage,
+        carryOverKey: uuid(),
+        carryOverTotal: 0,
+        stripeSubscriptionItemId: item.stripeSubscriptionItemId,
+    });
+}
+
+export const backfillSubscriptions = async () => {
```
Kitenite, line 104: @http-teapot , could you take a look at this backfill script? 

This will one-shot the entire backfill operation. Along with reading the data from Stripe (add to .env under packages/db). I'm running some tests but wanna make sure I'm calculating the rate-limit items correctly.

packages/db/src/seed/stripe/backfill-subscriptions.ts
```diff
@@ -0,0 +1,135 @@
...
+            stripeCurrentPeriodEnd,
+        });
+    }
+
+    return items;
+}
+
+const insertRateLimit = async (tx: PgTransaction<any, any, any>, item: StripeItem) => {
+    console.log(`Inserting rate limit for subscription ${item.subscription.id}`);
+    // One month from the start date. NOT SURE IF THIS IS CORRECT.
```
Kitenite: @http-teapot should we use a better end date here?

http-teapot: If `stripeCurrentPeriodEnd` is available then I'd use that so you don't have to calculate 31/30 days based on month

packages/db/src/seed/stripe/backfill-subscriptions.ts
```diff
@@ -0,0 +1,121 @@
...
+            stripeCurrentPeriodEnd,
+        });
+    }
+
+    return items;
+}
+
+const insertRateLimit = async (tx: PgTransaction<any, any, any>, item: StripeItem) => {
+    // One month from the start date. NOT SURE IF THIS IS CORRECT.
+    const endDate = new Date(item.subscription.startedAt.getTime() + 30 * 24 * 60 * 60 * 1000);
```
http-teapot: btw, `date-fns` has the function `addMonths`

```
import { addMonths } from "date-fns/addMonths";

addMonths(new Date(), 1));
```

avoids doing math *and* it handles the number of days in a month

February 1st + 1 month = March 1st
January 31st + 1 month = February 28th

https://date-fns.org/docs/addMonths

PR #2469, opened 2025-07-17
Fix decimal input support, add VH units.

Rish-it:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor


Kitenite (changes requested):
There seems to be a regression wherein the key up and down behavior to increment/decrement value no longer works. Could you check that @Rish-it ?

Rish-it:
@Kitenite Thanks for flagging! The issue was due to string state handled outside the hook. I've moved that logic into `useInputControl`, restoring arrow key behavior while keeping decimal support. Should be resolved now.

Kitenite:
And thanks for adding arrow key support for the range input

PR #2515, opened 2025-07-22
feat: page selector

spartan-vutrannguyen:
## Description

- Added a Pages controller to the left of the top bar — lets you go back, forward, and quickly switch between pages.
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

https://github.com/user-attachments/assets/1e6634c3-5474-440f-8e03-520e22cfe42a

## Additional Notes

----

> [!IMPORTANT]
> Introduces a page selector feature with navigation history management and URL inference for enhanced page navigation.
> 
>   - **Behavior**:
>     - Adds `PageSelector` component in `page-selector.tsx` for navigating between pages.
>     - Integrates `PageSelector` into `TopBar` in `top-bar.tsx`.
>     - Updates `WebFrameComponent` in `web-frame.tsx` to support navigation.
>   - **Navigation**:
>     - Implements `FrameNavigationManager` in `navigation.ts` for managing navigation history.
>     - Adds navigation methods `goBack`, `goForward`, and `navigateToPath` in `manager.ts`.
>     - Tests navigation functionality in `navigation.test.ts`.
>   - **Utilities**:
>     - Adds `inferPageFromUrl` in `urls.ts` to infer page details from URL.
>     - Tests `inferPageFromUrl` in `urls.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 90cf63e65808df780bfdec26ada1e41bfc66b59a. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/src/components/store/editor/pages/index.ts
```diff
@@ -20,6 +20,11 @@ export class PagesManager {
     private currentPath = '';
     private groupedRoutes = '';
     private _isScanning = false;
+    
+    // Navigation history
+    private navigationHistory: string[] = [];
```
Kitenite: This doesn't seem to take into account multiple frames. Not sure if this is even the right place to put this functionality instead of at the frame level. Please test with multiple frames. Might be better there since each frame will handle its own history.

apps/web/client/src/app/project/[id]/_components/canvas/frame/web-frame.tsx
```diff
@@ -146,40 +142,10 @@ export const WebFrameComponent = observer(
...
 
-            const syncMethods = {
-                supportsOpenDevTools: () =>
-                    !!iframe.contentWindow && 'openDevTools' in iframe.contentWindow,
-                setZoomLevel: (level: number) => {
-                    zoomLevel.current = level;
-                    iframe.style.transform = `scale(${level})`;
-                    iframe.style.transformOrigin = 'top left';
-                },
-                loadURL: (url: string) => {
```
Comment (spartan-vutrannguyen) [inline apps/web/client/src/app/project/[id]/_components/canvas/frame/web-frame.tsx:201]: @Kitenite is there any reason why we call updateAndSaveToStorage from iframe? I've updated to call it directly from FrameManager

Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/web-frame.tsx:201]: This is to make sure if a page changes URL, it is saved on the frame on database. Does the new method handle if the url is change from any source?

apps/web/preload/script/api/navigation.ts
```diff
@@ -0,0 +1,57 @@
+import { penpalParent } from '../index';
+
+let lastUrl = window.location.href;
+
+export function listenToNavigationChanges() {
```
Kitenite: Overriding a window object is something we REALLY should not be doing... This can create some real issues for users project. What's the reason for the override?

apps/web/client/src/components/store/editor/frames/manager.ts
```diff
@@ -0,0 +1,345 @@
...
+        const frameData = this.get(id);
+        if (!frameData?.view) {
+            console.error('Frame view not found for reload', id);
+            return;
+        }
+        frameData.view.reload();
+    }
+
+    // Navigation history methods
+    get canGoBack(): boolean {
```
Kitenite: Wouldn't this be wrong or incomplete when there are multiple frames selected? Why even have these methods? Or show incorrect information for the wrong frames that use this?

Why not let the frame themselves call navigationmanger.canGoBack directly?

apps/web/preload/script/api/navigation.ts
```diff
@@ -0,0 +1,57 @@
...
+    // Override replaceState
+    history.replaceState = function(...args) {
+        originalReplaceState.apply(history, args);
+        setTimeout(() => {
+            notifyNavigationChange();
+        }, 100);
+    };
+    
+    // Listen for popstate events
+    window.addEventListener('popstate', () => {
```
Kitenite: This did not work in my testing. Can you give an example of it working? Why the setTimeout?

apps/web/preload/script/api/navigation.ts
```diff
@@ -0,0 +1,57 @@
...
+        setTimeout(() => {
+            notifyNavigationChange();
+        }, 100);
+    });
+}
+
+function notifyNavigationChange() {
+    if (penpalParent) {
+        try {
+            const currentUrl = window.location.href;
```
Kitenite: I would assume global state is cleared when page is changed. Can you confirm that this actually works? Why not just look in the history object instead of saving as a variable?

apps/web/preload/script/api/navigation.ts
spartan-vutrannguyen: I understand that we really should not overriding but look likes there is no way to fully capture a navigation without overriding.

apps/web/client/src/app/project/[id]/_components/canvas/frame/web-frame.tsx
```diff
@@ -220,22 +188,45 @@ export const WebFrameComponent = observer(
...
+            }
+
             // Register the iframe with the editor engine
             editorEngine.frames.registerView(frame, iframe as WebFrameView);
 
             return Object.assign(iframe, {
                 ...syncMethods,
                 ...remoteMethods,
-            }) satisfies WebFrameView;
+            }) as WebFrameView;
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/canvas/frame/web-frame.tsx]: Please use satisfies when possible. forcing type can cause runtime errors

Kitenite (approved):
I fixed the regression by making it possible to click on the folder and select it at the same time. It's not perfect and I'm not a fan of how the pages tab determines active pages (way before this PR). 

But works well, thanks!

PR #2610, opened 2025-08-04
fix: visual improvement to the frame

### LINKED ISSUE #2609: [bug] Visual improvements to the frame / window controls
Some of these are left over and not addressed from #2506 

Canvas
1. When the user has the canvas selected, then they use the handles, the canvas is unselected (as in, the top toolbar disappears, and the header color turns grey when it should stay teal)

Image

2. Persist the window controls between Design and Preview modes. So, the user can have controls to adjust and move the selected canvas around.

Image

3. When the frame is selected, the border should also be teal, not grey. 

Image

4. These buttons in the toolbar need to have a rounded class. It's SUPER subtle, but important for consistency.

Image

Image

5. When you select a new device size from the frame controls, it deselects the frame and you have to click it again. Make it so you can change the devices and keep the frame selected. 

Image

6. Use the Rotate angle icon used in the gradient color picker instead of the "history clock" icon for the icon in the window controls bar

Image

Image

spartan-vutrannguyen:
## Related Issues

- close #2609 
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

https://github.com/user-attachments/assets/fdcd1c10-416f-4dec-9321-57bb385cd7f0

### DIFF
--- a/apps/web/client/src/components/store/editor/frames/manager.ts
+++ b/apps/web/client/src/components/store/editor/frames/manager.ts
@@ -68,7 +68,8 @@ export class FramesManager {
     }
 
     registerView(frame: Frame, view: WebFrameView) {
-        this._frameIdToData.set(frame.id, { frame, view, selected: false });
+        const isSelected = this.isSelected(frame.id);
+        this._frameIdToData.set(frame.id, { frame, view, selected: isSelected });
         const framePathname = new URL(view.src).pathname;
         this._navigation.registerFrame(frame.id, framePathname);
     }
@@ -106,6 +107,7 @@ export class FramesManager {
     }
 
     private notify() {
+        console.log('notify', this._frameIdToData);
         this._frameIdToData = new Map(this._frameIdToData);
     }
 
@@ -248,7 +250,7 @@ export class FramesManager {
         const existingFrame = this.get(frameId);
         if (existingFrame) {
             const newFrame = { ...existingFrame.frame, ...frame };
-            this._frameIdToData.set(frameId, { ...existingFrame, frame: newFrame });
+            this._frameIdToData.set(frameId, { ...existingFrame, frame: newFrame, selected: existingFrame.selected });
         }
         await this.saveToStorage(frameId, frame);
     }
--- a/apps/web/client/src/components/store/editor/frames/manager.ts
+++ b/apps/web/client/src/components/store/editor/frames/manager.ts
@@ -107,7 +107,6 @@ export class FramesManager {
     }
 
     private notify() {
-        console.log('notify', this._frameIdToData);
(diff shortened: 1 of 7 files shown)


Kitenite:
Looks good! Can you please test with multiple frames? Seems like selecting multiple frames is not showing up as intended. I would expect both to be highlighted in this case
 
Screenshot 2025-08-04 at 4 38 37 PM

apps/web/client/src/app/project/[id]/_components/canvas/frame/gesture.tsx
```diff
@@ -61,7 +61,7 @@ export const GestureScreen = observer(({ frame, isResizing }: { frame: WebFrame,
                         break;
                     case MouseAction.MOUSE_DOWN:
                         if (el.tagName.toLocaleLowerCase() === 'body') {
-                            editorEngine.frames.select([frame]);
+                            editorEngine.frames.select([frame], !e.shiftKey);
```
Comment (Kitenite) [inline apps/web/client/src/app/project/[id]/_components/canvas/frame/gesture.tsx]: I'd recommend a parameter where you don't have to pass the opposite param. 
Something like multiselect = false will allow passing `e.shiftKey` instead of `!e.shiftKey` and make it easier to understand

apps/web/client/src/components/store/editor/frames/manager.ts
```diff
@@ -85,8 +86,10 @@ export class FramesManager {
         return this._frameIdToData.get(id)?.selected ?? false;
     }
 
-    select(frames: Frame[]) {
-        this.deselectAll();
+    select(frames: Frame[], exclusive = true) {
+        if (exclusive) {
```
Kitenite: Since we're here can we instead of exclusive makes it something like multiselect = false and handle the case where it's already selected to unselect it?

https://github.com/onlook-dev/onlook/blob/main/apps/web/client/src/components/store/editor/element/index.ts#L54

apps/web/client/src/app/project/[id]/_components/canvas/frame/gesture.tsx
Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/canvas/frame/gesture.tsx]: @spartan-vutrannguyen just updated, I meant to say `multiselect = false` since omitting == false is more expected behavior. So that the default is a regular select. Thanks

apps/web/client/src/components/store/editor/frames/manager.ts
Kitenite: Updated this comment as well

Kitenite:
Still did not address selecting multiple elements should show multiple pages as selected. I will merge becauseI don't want the PR to go stale but please follow up with a PR to address that when you have time. Thanks.

Screenshot 2025-08-05 at 12 00 39 PM

PR #2613, opened 2025-08-04
feat: implement comprehensive edit tools toolset

Kitenite:
## Summary
- Implement comprehensive editing tools for enhanced AI integration
- Add Bash, Glob, Grep, LS, Read, Edit, MultiEdit, Write tools
- Add Notebook manipulation tools (Read/Edit)
- Add Web tools (Fetch/Search) and utility tools (TodoWrite, ExitPlanMode)

## Key Features
✅ **Sandbox Integration** - All file operations use `editorEngine.sandbox` instead of direct Node.js APIs  
✅ **Type Safety** - Full TypeScript support with Zod schemas for validation  
✅ **Error Handling** - Comprehensive error handling for all operations  
✅ **Integration** - Seamlessly integrated with existing Onlook tools system  
✅ **Build Success** - All TypeScript compilation errors resolved  

## Test plan
- [x] TypeScript compilation passes
- [x] Build completes successfully
- [x] All tools properly integrated with existing system
- [x] Sandbox operations work correctly
- [x] Manual testing of individual tools
- [x] Integration testing with AI workflows

🤖 Generated with [Claude Code](https://claude.ai/code)

----

> [!IMPORTANT]
> Implements 15 editing tools for AI integration with sandboxed file operations, error handling, and modular tool handling.
> 
>   - **Behavior**:
>     - Implements 15 editing tools for AI integration, including `Task`, `Bash`, `Glob`, `Grep`, `LS`, `Read`, `Edit`, `MultiEdit`, `Write`, `Notebook`, `Web`, `TodoWrite`, `ExitPlanMode`.
>     - Uses `editorEngine.sandbox` for file operations instead of direct Node.js APIs.
>     - Comprehensive error handling and TypeScript support with Zod schemas.
>   - **Tool Handling**:
>     - Refactors tool handling into separate modules: `cli.ts`, `edit.ts`, `guide.ts`, `plan.ts`, `read.ts`, `sandbox.ts`, `web.ts`.
>     - Introduces `toolset.ts` to define `ASK_TOOL_SET` and `BUILD_TOOL_SET`.
>   - **Code Changes**:
>     - Renames `askToolSet` to `ASK_TOOL_SET` and `buildToolSet` to `BUILD_TOOL_SET` in `route.ts`.
>     - Removes `tools.ts` and replaces it with modular handlers.
>     - Adds new handlers for CLI, edit, guide, plan, read, sandbox, and web tools.
[...]

PR #2622, opened 2025-08-06
Code provider abstraction

http-teapot:
## Description

This PR creates a new package `@onlook/code-provider` and abstracts any CodeSandbox away from the main codebase.

This is a first iteration with the sole goal of abstracting CodeSandbox while refactoring very little the main code. It has two added benefits:

1. Reduces the amount of code to review.
2. Helps with regression testing.
3. Avoids premature optimization when things could drastically change once we start integrating other providers.

The providers are designed to be singletons; makes sense because it tends to do authenticated requests and also real-time updates via websockets (I think).

The functions used for AI tools and inner code have been merged into one function; we could technically surface more AI tools as a result.

Most CodeSandbox functions (except the real-time events) could run on the server; when the provider is instantiated it knows how to initialize itself based on the environment.

Side note: how the provider is instantiated in this PR is rudimentary but in the future we can store the provider somewhere in the database.

A provider is defined by interfaces and abstract classes, and each provider is self-contained into its own folder. It should help with contributions – although, as I mention above, the methods in the interfaces/classes will change as we start integrating other providers.

For bullet-proofing every method takes an input as object (even if empty) and return an object (even if empty).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [x] Refactor

## Testing

1. Created a new project.
2. Created several prompts.
3. Replayed prompts.
4. Used the terminal.
5. Tried to publish projects but I think I need a key.

## Additional Notes

----

> [!IMPORTANT]
> Introduces `@onlook/code-provider` package to abstract code sandboxing, implementing providers for CodeSandbox and NodeFs.
> 
>   - **New Package**:
>     - Introduces `@onlook/code-provider` for code sandboxing abstraction.
>     - Implements `Provider` interface for `CodeSandbox` and `NodeFs`.
>   - **CodeSandbox Provider**:
>     - Implements methods for file operations, terminal management, and project handling in `codesandbox/index.ts`.
>     - Utilizes `@codesandbox/sdk` for interactions.
>   - **NodeFs Provider**:
>     - Provides a basic structure with unimplemented methods in `nodefs/index.ts`.
>   - **Types and Interfaces**:
>     - Defines types and interfaces in `types.ts` for provider operations.
>     - Includes file operations, terminal commands, and project management.
>   - **Utility Functions**:
>     - Adds utility functions for file operations in `codesandbox/utils`.
>   - **Miscellaneous**:
>     - Updates `package.json` to include new dependencies and scripts.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for e799598ca948664489599265176903e1b016820f. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


apps/web/client/package.json
```diff
@@ -34,8 +34,8 @@
         "@codemirror/lang-javascript": "^6.2.3",
         "@codemirror/lang-json": "^6.0.1",
         "@codemirror/lang-markdown": "^6.1.7",
-        "@codesandbox/sdk": "^1.1.6",
```
http-teapot, line 37: finger-guns-michael-scott-3753599043

apps/web/client/src/app/project/[id]/_components/left-panel/image-tab/hooks/use-image-move.ts
```diff
@@ -63,11 +68,6 @@ export const useImageMove = () => {
                 return;
             }
 
-            const session = editorEngine.sandbox.session?.session;
-            if (!session) {
-                throw new Error('No sandbox session available');
-            }
-
```
Comment (http-teapot) [inline apps/web/client/src/app/project/[id]/_components/left-panel/image-tab/hooks/use-image-move.ts:70]: this was redundant, it is checked inside `editorEngine.sandbox.copy(currentPath, fullPath, true);` below

there are several places in the PR that remove checks like this

apps/web/client/src/components/store/editor/sandbox/index.ts
```diff
@@ -310,14 +336,20 @@ export class SandboxManager {
...
-            const { downloadUrl } = await this.session.session.fs.download('./');
+            const { url } = await this.session.provider.downloadFiles({
+                args: {
+                    path: './',
+                },
+            });
             return {
-                downloadUrl,
+                // in case there is no URL provided then the code must be updated
+                // to handle this case
```
http-teapot, line 357: Perhaps, in the future, `downloadFiles` may return the files as binary instead of a URL.

apps/web/client/src/server/api/routers/publish/helpers/fork.ts
```diff
@@ -1,20 +1,40 @@
...
+import { CodeProvider, createCodeProviderClient, type Provider } from '@onlook/code-provider';
 
-const sdk = new CodeSandbox(env.CSB_API_KEY);
+export async function forkBuildSandbox(
+    sandboxId: string,
+    userId: string,
+    deploymentId: string,
+): Promise<{ provider: Provider; sandboxId: string }> {
+    // not a big fan of this pattern... may need to refactor `createProject` into a static method
+    // though, that entirely depends on other provider implementations
```
http-teapot, line 9: The original code did a simple function call `sdk.sandboxes.create` whereas the new pattern goes through an instantiated class.

We could do a static function on the class, I think. I prefer to wait until we are integrating other providers to refactor this.

And perhaps, the entire function `forkBuildSandbox` will change instead.

packages/code-provider/src/providers/nodefs/index.ts
http-teapot, line 1: just an empty provider that implements the necessary classes, could create a mock provider to test behavior in the future

apps/web/client/package.json
Kitenite: Woot!

apps/web/client/src/components/store/editor/pages/helper.ts
```diff
@@ -998,10 +1000,7 @@ export const addSetupTask = async (sandboxManager: SandboxManager) => {
         },
     };
     const content = JSON.stringify(tasks, null, 2);
-    await sandboxManager.writeFile(
-        './.codesandbox/tasks.json',
-        content,
-    );
+    await sandboxManager.writeFile('./.codesandbox/tasks.json', content);
```
Kitenite, line 1003: This may need to be refactored as well for the provider

packages/code-provider/src/providers/codesandbox/utils/create-file.ts
```diff
@@ -0,0 +1,20 @@
+import { WebSocketSession } from '@codesandbox/sdk';
+import { fileExists, writeFile } from './utils';
+import type { CreateFileInput, CreateFileOutput } from '../../../types';
+
+export async function createFile(
```
Kitenite: This tool was deprecated. 

Would it make sense to decouple the tool-specific logic from the actual provider implementation? Seems like this function for example would have the same logic across providers.

https://github.com/onlook-dev/onlook/pull/2613/files#r2258144514

packages/code-provider/src/providers/codesandbox/utils/edit-file.ts
```diff
@@ -0,0 +1,64 @@
+import { WebSocketSession } from '@codesandbox/sdk';
+import { getBaseName, getDirName, normalizePath } from '@onlook/utility';
+import { type SandboxFile } from '@onlook/models';
+import { readRemoteFile, writeFile } from './utils';
+import type { EditFileInput, EditFileOutput } from '../../../types';
+
+export async function editFile(
```
Kitenite: Edit file tools also changed. 

https://github.com/onlook-dev/onlook/pull/2613/files#r2258144514

I think perhaps we should keep tools logic in the tool section and pass in the generic provider for read and write instead. Similar to comment above.

Kitenite (commented):
On the note of tools, there seems to be some type issues here with the new tools params

https://github.com/onlook-dev/onlook/blob/b21bf120c4b5d3d28362b7d8a8f67734668be11a/packages/code-provider/src/types.ts#L1

You can confirm with `cd apps/web/client && bun run typecheck`

apps/web/client/src/components/store/editor/pages/helper.ts
http-teapot: ah good catch!

yeah the whole terminal/task layer of abstraction is a big TBD atm, not sure how that's going to translate in other providers

packages/code-provider/src/providers/codesandbox/utils/create-file.ts
http-teapot: I am not sure I follow, `createFile` would be called inside the function you linked at line 34, the reason it's structured that way is because I copy/pasted the existing code and inside the `createFile` function it calls a function `writeFile` which is also called when `editFile`. Right now, the provider reflects the current code implementation but I imagine in the future it'll change to be more streamlined.

apps/web/client/src/server/api/routers/publish/helpers/fork.ts
Kitenite: Perhaps they can all expose a fork function, since it wouldn't make sense to fork a provider into a different provider.

packages/code-provider/src/providers/codesandbox/utils/create-file.ts
Kitenite: I see, I conflated this implementation to the `handleCreateFileTool`. This makes sense + the convo we had in slack

packages/code-provider/src/providers/codesandbox/utils/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+              }
+            : {
+                  type,
+                  path: filePath,
+                  content: content as string,
+              };
+    return newFile;
+}
+
+export function isJsxFile(filePath: string): boolean {
```
Kitenite: This is a non-generic functionality that we'll use for all the providers

packages/code-provider/src/providers/codesandbox/utils/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+    const extension = path.extname(filePath);
+    if (!extension || !NEXT_JS_FILE_EXTENSIONS.includes(extension)) {
+        return false;
+    }
+    return true;
+}
+
+export async function writeFile(
```
Kitenite: This seems to be a duplicate of the sandbox higher level writeFile functionality. I think we'll keep the processingFileForMapping and isJsxFile logic at the higher level since it's more app specific

https://github.com/onlook-dev/onlook/blob/ecbc86f74aae6099b012d898750cbe2294c2fe0c/apps/web/client/src/components/store/editor/sandbox/index.ts#L232

http-teapot:
> $ bun run typecheck

Only returns

> $ tsc --noEmit

Was the issue fixed? Am I doing something wrong?

Kitenite:
> > $ bun run typecheck
> 
> Only returns
> 
> > $ tsc --noEmit
> 
> Was the issue fixed? Am I doing something wrong?

That usually means no type issues!

packages/code-provider/src/providers/codesandbox/utils/read-file.ts
```diff
@@ -0,0 +1,37 @@
...
+            },
+        };
+    } else {
+        // const base64Content = file.content ? convertToBase64(file.content) : '';
+        return {
+            file: {
+                path: file.path,
+                content: file.content,
+                type: file.type,
+                toString: () => {
```
Kitenite, line 31: Curious what the advantage of using textdecoder here vs the convertToBase64 method. These give different results of course but want to give your take on how we should handle getting base64 vs text decoded format. 

```
new TextDecoder().decode(file.content) 
```

apps/web/client/src/server/api/routers/publish/helpers/fork.ts
Kitenite: Ah I see what you mean now. Each provider is tied to a single sandbox... So it acts more like a sandbox session than a "provider".

Perhaps we can have certain static methods exposed (in this case create) where it's not through instantiation? 
`CodeProvider.CodeSandbox.create()`. But can defer

packages/code-provider/src/providers/codesandbox/utils/read-file.ts
```diff
@@ -0,0 +1,37 @@
...
+                    return file.content ? new TextDecoder().decode(file.content) : '';
```
Kitenite: Going to update this to base64 for now since we want to retain same functionality

packages/code-provider/src/providers/codesandbox/utils/utils.ts
```diff
@@ -0,0 +1,40 @@
+import type { WebSocketSession } from '@codesandbox/sdk';
+import { type SandboxFile } from '@onlook/models';
+import { isImageFile } from '@onlook/utility';
+
+export function getFileFromContent(filePath: string, content: string | Uint8Array) {
```
Kitenite, line 5: These can almost be removed as they're only used in 1 place but will save for another time

packages/code-provider/src/types.ts
```diff
@@ -0,0 +1,280 @@
...
+}
+
+export interface PauseProjectInput {}
+export interface PauseProjectOutput {}
+
+export interface StopProjectInput {}
+export interface StopProjectOutput {}
+
+export interface ListProjectsInput {}
+export interface ListProjectsOutput {}
```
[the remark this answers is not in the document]
Kitenite: This seems correct. Is the intention to wait and see how other providers might do before settling on a definition here @http-teapot ?

PR #2666, opened 2025-08-13
fix: handle zoom in preview

spartan-vutrannguyen:
## Description

- Currently, zooming on a web frame while in preview mode zooms the entire canvas

https://github.com/user-attachments/assets/c48a08f6-36fc-4621-9548-a9c614367fe7

- Added a handler to handle zoom action correctly

https://github.com/user-attachments/assets/e62cab1a-ebce-4377-98d7-62e957bd9d2b

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes

----

> [!IMPORTANT]
> Fix zoom handling in preview mode by adding a wheel event handler to forward events to the canvas container and updating event handling logic.
> 
>   - **Behavior**:
>     - Adds `onFrameWheel` handler in `web-frame.tsx` to forward wheel events to the canvas container.
>     - Introduces `listenForWheelZoom` in `wheel.ts` to intercept wheel events with `ctrlKey` or `metaKey` and prevent default behavior if not handled inside.
>   - **Event Handling**:
>     - Updates `PenpalParentMethods` in `parent.ts` to include `onFrameWheel`.
>     - Adds `listenForWheelZoom` to `index.ts` to start listening for wheel events.
>   - **Misc**:
>     - Removes global wheel event listener in `main.tsx` to prevent unintended interception.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 213224392f2c25a82c1f72a6234b3303ff94624b. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## Summary by CodeRabbit

* **New Features**
  * Ctrl/⌘ + mouse-wheel zoom now works when scrolling over embedded frames — wheel input is forwarded so zoom targets the outer canvas and centers on the cursor.
  * Wheel-zoom listening is initialized as part of DOM event wiring for consistent behavior.

* **Bug Fixes**
  * Stopped globally intercepting Ctrl/⌘ wheel events; native scrolling/zoom now applies for scrollable, transformed, or media elements.
  * Layout overflow constrained to avoid unintended scrolling.


Kitenite:
This is great!

Can we also try this: make it so that you can scroll but if you over-scroll on the iframe, you just scroll the canvas instead? I just want to see how that feels since a lot of people got confused when they can't scroll in preview mode. 

Or if you're scrolling inside the canvas, let's just have an indication since we're now detecting it.

Kitenite:
One more bug and you can see it from the preview video, the mouse doesn't line up correctly with the mouse outside. If this is not fixable, I think it's ok to do nothing on zoom which is still an improvement. 

Be sure to test this with projects that have some zoom functionality inside of it like a Figma clone.

Issue #2504, opened 2025-07-21
[bug] Line-breaks (SHIFT+ENTER) aren't writing to code – multi-line editing


drfarrell:
#### Describe the bug

Tried SHIFT+ENTER for the "In the heart of SF" text but it doesn't actually write to code. The linebreak happens in the view, but it doesn't reflect in the code. 

<img width="1522" height="848" alt="Image" src="https://github.com/user-attachments/assets/25930bce-fbbf-4dcd-b4dc-784018549381" />

The greater challenge though is allowing for multi-line editing. So, if a user has a line break or if text already comes with a <br /> in it, we can edit it and remove the break if we want.

Paribesh01:
@drfarrell can I work on this ?

rohan-chidurala:
@Paribesh01 i found the core issue but like its still kinda buggy on the frontend like the current frame you are on bugs out for some reason i couldn't really figure out why?

PR #2809, opened 2025-09-06
fix: handle breaks in tags

rohan-chidurala:
## Description

Update the preload script so it can detect <br> elements with attributes (e.g., <br data-oid="">).

Right now, it’s still buggy when adding multiple line breaks. From what I can tell, the issue happens during sandbox updates/refreshes. The data and file changes look correct, but when the sandbox refreshes, only the current frame you’re working on glitches—the other frames are fine. I couldn’t fully figure out why, so someone else might need to dig deeper.

## Related Issues

#2504 

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor


Kitenite:
will take a look thanks @rohan-chidurala !

PR #2925, opened 2025-09-26
fix: branch and window dragging reference issues

### LINKED ISSUE #2908: [bug] Branch and window dragging use the wrong references
#### Describe the bug

This has happened in two cases: 
1) When a user creates a new branch and then tries to drag the webframe topbar to arrange it, it acts as if the user is dragging the original branch

2) When the user has one window selected and tries to drag another adjacent window, it pretends like the first window is the one still in focused. 

Image

The expected behavior would be that the user can click on any webframe and it would select and drag that topframe. Right now it seems to keep the first frame selected and because the mouse hovers over a topbar area it enables the ability to drag / move the frame.

DavidReque:
## Description

Fix incorrect frame dragging references after branch changes or when dragging an adjacent window.
- TopBar: on mousedown, if not multiselect (no Shift) and the clicked frame isn’t selected, select it first, then start drag with the fresh selection; ignore right‑click; prevent default.
- FramesManager.select: single select now replaces selection (no toggle); Shift‑multiselect keeps toggle behavior.

## Related Issues

Fixes #2908

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Release
- [ ] Refactor

## Additional Notes
- Clicking any frame’s topbar now immediately selects and drags that specific frame.
- After switching branches, dragging affects frames from the current branch only.

----

> [!IMPORTANT]
> Fixes frame selection and dragging issues by updating selection logic and handling mouse events in `index.tsx` and `manager.ts`.
> 
>   - **Behavior**:
>     - In `index.tsx`, `handleMouseDown` now prevents default actions on right-clicks and selects the frame if not already selected before dragging.
>     - Dragging affects frames from the current branch only after branch changes.
>     - Clicking a frame's top bar selects and drags that frame immediately.
>   - **Selection Logic**:
>     - In `manager.ts`, `select()` method updated to replace selection instead of toggling when not in multiselect mode.
>     - Shift+click multiselect and group drag behavior remains unchanged.
>   - **Misc**:
>     - Fixes issue #2908 related to incorrect frame dragging references.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for a2175b7fb71d8774a8fa27c0c8b87077d0c1a7be. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Dragging frames from the top bar now only responds to the primary mouse button and no longer triggers unwanted text selection.
  * Clicking a frame (without Shift) reliably selects it before moving, preventing accidental moves of previously selected frames.
  * Single-click selects (rather than toggles) in single-selection mode; Shift-click still toggles for multi-select.
  * Moving multiple frames now reliably uses the current selection when available.


Kitenite:
This is awesome, thanks @DavidReque !

PR #2977, opened 2025-10-08
Separate out images from the rest of the context, don't clear it as user clicks in canvas

saddlepaddle:
## Description
Screenshot 2025-10-08 at 2 00 22 PM

1) Pulls out the image context into it's own separate context both in mobx and in the UI
2) Avoid clearing the image when the user clicks around
3) Sends images directly to public/ by default instead of a folder in public

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Summary by CodeRabbit

- New Features
  - Drag-and-drop, paste and screenshot support for inserting images into chat with local vs external source distinction.
  - Prompts now list local images separately from available external images and include upload guidance.

- Improvements
  - Improved drag-state visuals and responsive chat input UI.
  - Image handling uses data-URL encoding for more reliable previews and preserves images when editing context.
  - Upload tool guidance clarified to avoid re-uploading local images.

- Bug Fixes
  - Image menu closes after adding an image; clearer error messages.

----

> [!IMPORTANT]
> This PR enhances image handling by separating image contexts, improving drag-and-drop functionality, and refining image upload processes across the application.
> 
>   - **Behavior**:
>     - Separates image context from other contexts in `ChatContext` in `context.ts`.
>     - Prevents image context clearing on canvas click in `context.ts`.
>     - Images are now uploaded directly to `public/` by default in `upload-image.ts`.
>   - **UI Components**:
>     - Renames `DraftImagePill` to `ImagePill` and updates references in `create.tsx` and `input-context-pills.tsx`.
>     - Adds drag-and-drop support for images in `chat-input/index.tsx` and `image-item.tsx`.
>     - Updates `ImagePill` to handle image removal in `image-pill.tsx`.
>   - **Utilities**:
>     - Introduces `convertToBase64DataUrl` in `file.ts` for consistent image encoding.
>     - Updates `read-file.ts` to use `convertToBase64` for binary files.
>   - **Models**:
>     - Adds `source` attribute to `ImageMessageContext` in `context.ts` to distinguish between local and external images.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=onlook-dev%2Fonlook&utm_source=github&utm_medium=referral)<sup> for 66e76b74a5d139e16e67d281333f48d11295c38b. You can [customize](https://app.ellipsis.dev/onlook-dev/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-input/index.tsx
+++ b/apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-input/index.tsx
@@ -155,6 +155,9 @@ export const ChatInput = observer(
             try {
                 await onSendMessage(savedInput, chatMode);
                 setInputValue('');
+                setTimeout(() => {
+                    onScrollToBottom();
+                }, 0);
             } catch (error) {
                 console.error('Error sending message', error);
                 toast.error('Failed to send message. Please try again.');


apps/web/client/src/components/store/editor/chat/context.ts
```diff
@@ -30,7 +31,13 @@ export class ChatContext {
             }),
             (
                 { elements, frames },
-            ) => this.generateContextFromReaction({ elements, frames }).then((context) => (this.context = context)),
+            ) => this.generateContextFromReaction({ elements, frames }).then((context) => {
+                // Preserve existing images when updating context from canvas selection
```
Kitenite: @saddlepaddle why do we need the local image context at all if we just do this?

packages/models/src/chat/message/context.ts
```diff
@@ -41,6 +42,13 @@ export type ImageMessageContext = BaseMessageContext & {
     id?: string;
 };
 
+export type LocalImageMessageContext = BaseMessageContext & {
```
Kitenite: Should we actually update the file context type to handle binary/image instead of making a type specifically for local image?

apps/web/client/src/components/store/editor/chat/context.ts
saddlepaddle: hmm well we need it so the LLM knows it's a local file basically - that way it knows how to grab it / what to add to the dom / that it shouldn't try uploading it.

packages/models/src/chat/message/context.ts
saddlepaddle: ooo good call! yeah we can merge 👍🏼. It's still useful to differentiate the two though since we want the agent to know to not try to re-upload local assets.

apps/web/client/src/components/store/editor/chat/context.ts
saddlepaddle: I did it so the agent is aware of local vs. not local files! We have to upload non-local files first before we can use them, but for local files we don't want the llm to re-upload them each time it tries to use it 👍🏼

Kitenite:
@saddlepaddle this rebase seems to have override changes from main under `web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-input/index.tsx`can you double check?

apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-input/index.tsx
```diff
@@ -36,389 +39,428 @@ interface ChatInputProps {
...
+                }
             }
-        } else if (e.key === 'Enter' && !e.shiftKey && !isComposing) {
-            e.preventDefault();
-            e.stopPropagation();
+        };
 
-            if (suggestionRef.current?.handleEnterSelection()) {
-                setTimeout(() => textareaRef.current?.focus(), 0);
+        async function sendMessage() {
```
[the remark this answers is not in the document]
Comment (Kitenite) [reply apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-input/index.tsx:149]: @saddlepaddle this is a valid regression which makes me worried about what else is removed.

PR #2986, opened 2025-10-09
Fix: iframe and sandbox retry mechanism

saddlepaddle:
## Description

1) Have the script injection wait for the sandbox provider to be ready
2) Add extra retry logic for starting up a sandbox (we were still failing even with the server-side retries)
3) Allows the sandbox to still be accessible if the preload script injection fails (probably better than fully bricking the product)
4) (main issue) Prevent SessionManager from starting before sandboxmanager is ready

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Summary by CodeRabbit

* **Bug Fixes**
  * Clear failure feedback when the project frame can’t connect, avoiding silent errors.
  * More accurate preload readiness detection to reduce stuck loading overlays.
  * Automatic retry for sandbox connections to minimize transient connection failures.
  * Multi-branch restore: removed misleading success toast on partial success; shows an error if any restore fails.

* **Refactor**
  * Streamlined internal state management for preload and connection flow to improve reliability.

### DIFF
--- a/apps/web/client/src/components/store/editor/sandbox/index.ts
+++ b/apps/web/client/src/components/store/editor/sandbox/index.ts
@@ -12,13 +12,17 @@ import { detectRouterConfig } from '../pages/helper';
 import { copyPreloadScriptToPublic, getLayoutPath as detectLayoutPath } from './preload-script';
 import { SessionManager } from './session';
 
+export enum PreloadScriptState {
+    NOT_INJECTED = 'not-injected',
+    LOADING = 'loading',
+    INJECTED = 'injected'
+}
 export class SandboxManager {
     readonly session: SessionManager;
     readonly gitManager: GitManager;
     private providerReactionDisposer?: () => void;
     private sync: CodeProviderSync | null = null;
-    preloadScriptInjected: boolean = false;
-    preloadScriptLoading: boolean = false;
+    preloadScriptState: PreloadScriptState = PreloadScriptState.NOT_INJECTED
     routerConfig: RouterConfig | null = null;
 
     constructor(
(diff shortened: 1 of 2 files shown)


apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-messages/multi-branch-revert-modal.tsx
```diff
@@ -71,25 +73,14 @@ export const MultiBranchRevertModal = ({
...
-                    `Successfully restored ${successCount} branch${successCount > 1 ? 'es' : ''}`,
-                    {
-                        description:
-                            failCount > 0
-                                ? `${failCount} branch${failCount > 1 ? 'es' : ''} failed to restore`
-                                : undefined,
-                    },
-                );
-            } else if (failCount > 0) {
+            if (failCount > 0) {
```
Comment (saddlepaddle) [inline apps/web/client/src/app/project/[id]/_components/right-panel/chat-tab/chat-messages/multi-branch-revert-modal.tsx:76]: unrelated cleanup change - we had too many toasts

apps/web/client/src/components/store/editor/sandbox/index.ts
```diff
@@ -100,6 +109,9 @@ export class SandboxManager {
             this.preloadScriptInjected = true;
         } catch (error) {
             console.error('[SandboxManager] Failed to ensure preload script exists:', error);
+            // Mark as injected to prevent blocking frames indefinitely
+            // Frames will handle the missing preload script gracefully
+            this.preloadScriptInjected = true;
```
Kitenite: should we do this in finally instead of doing both in try and catch?

Kitenite: in this case we should just remove this flag or merge it with the preloadScriptLoading flag

PR #3096, opened 2026-04-16
fix(mobile-preview): live edits propagate to Expo Go without restart

epinnock:
## Summary

- The runtime `AppRegistry` shim's closure-scoped `appRegistered` flag latched to `true` on the first push and silently dropped every later `registerComponent('main', ...)` call. The IIFE wrapper's fallback then threw `Entry module 'index.ts' did not call AppRegistry…` on every push after the first, so live edits never reached the device — Expo Go had to be killed and relaunched between every change.
- Removing the latch lets each push re-mount via `renderApp`. Combined with the prior commits on this branch (CustomEvent backup signal, Hermes class-decl boot fix, AppRegistry/IIFE flag bridge), the editor → bundle → push → render pipeline is now stable across repeated edits.
- Also gitignores `runtime/bundle.meta.json`, generated alongside the already-ignored `runtime/bundle.js`.

Validated end-to-end on iPhone 15 Simulator with SDK 54 Expo Go: six consecutive live edits propagated without restarting Expo Go.

## Test plan

- [ ] Local: `bun run typecheck`
- [ ] Local: `bun run test` in `packages/mobile-preview`
- [ ] Manual: open editor, scan QR with Expo Go (or boot iOS simulator with the manifest URL), make ≥3 successive edits to a `.tsx` and confirm each renders without relaunching Expo Go
- [ ] Manual: kill Expo Go mid-session, relaunch, confirm late-join replay delivers the latest cached push

🤖 Generated with [Claude Code](https://claude.com/claude-code)


epinnock:
Re-targeting to the epinnock fork.

PR #3121, opened 2026-06-29
perf(rls): wrap auth.uid()/current_setting() in (select …) for per-statement evaluation

dmitrymaranik:
Postgres re-evaluates a bare `auth.uid()` / `current_setting(…)` in an RLS policy **once per row**. Wrapping it in a scalar subquery — `(select auth.uid())` — makes it an InitPlan the planner evaluates **once per statement** and caches. It's predicate-equivalent (row visibility unchanged) and the optimization [Supabase documents](https://supabase.com/docs/guides/database/postgres/row-level-security) for RLS.

This adds a migration that wraps the 16 affected per-user policies:

- **Predicate-equivalent** — row visibility is unchanged; only *when* the auth call is evaluated changes (once per statement vs once per row).
- Covers **both `USING` and `WITH CHECK`** (the write path runs per inserted/updated row too).
- Verified with [pgrls](https://github.com/pgrls/pgrls) (an open-source RLS linter) on Postgres 16: the PERF001 findings clear to 0 after the change, nothing else altered.
- The SQL is `pgrls fix --rule PERF001` output, added as a new migration (`0020_wrap_rls_perf_initplan.sql`).

Happy to adjust the migration filename/placement to your conventions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission-check performance across key areas of the app, reducing repeated authentication lookups so actions feel faster and more responsive.
  * Optimized access control evaluation for file transfer and project-related workflows, including invitations, canvases, user settings, and user profile access.


dmitrymaranik:
Good catch — you're right that the EXISTS-nested `auth.uid()` calls were still bare (the tool I used, `pgrls`, wraps top-level policy calls only). I've updated the migration to wrap **every** `auth.uid()` call, including the ones nested inside correlated `EXISTS` subqueries (and the policies that had *only* nested calls). It's now 21 `ALTER POLICY` rewrites; verified on Postgres 16 that **0 bare `auth.*()` calls remain** in any policy predicate, with no double-wrapping. Still predicate-equivalent — row visibility is unchanged.

dmitrymaranik:
Good catch, and agreed it's out of scope for this migration (as the bot notes). This change wraps the policy predicates' direct `auth.uid()` calls — which is exactly what the InitPlan optimization targets and what the tool I used rewrites.

The `user_has_project_access(...)` helper calling `auth.uid()` internally is a separate surface — a `SECURITY DEFINER` function body, not a policy predicate — so a mechanical predicate-wrap doesn't (and shouldn't) touch it. Stabilizing that helper (passing the uid in, or caching it once inside the function) is a worthwhile follow-up for the maintainers, but independent of this migration. The InitPlan win on the direct calls stands.

